Issue #6, opened 2024-02-23
Mirror select dialog doesn't set a default value


Plootie:
Pretty much the title, should be a 2 second change

artehe:
Settinga default value was about 2 seconds but did give me the thought of should we leave this as created in code or should we actually make a view for it?

PR #9, opened 2024-02-23
Merge dialog views

### LINKED ISSUE #7: Merge the version selection dialogs
SelectSitVersionDialog and SelectServerVersionDialog are 99% identical so can just make one version selection dialog and pass the required data into them.

artehe:
We only need the one dialog view for version selection as they're nigh identical

Should resolve #7


artehe:
Also should fix #6 now too as it was somewhat related.

SIT.Manager.Avalonia/Services/InstallerService.cs
```diff
@@ -278,6 +287,76 @@ public class InstallerService(IBarNotificationService barNotificationService,
...
+
+            List<GithubRelease> result = [];
+            if (githubReleases.Any()) {
+                foreach (GithubRelease release in githubReleases) {
+                    var zipAsset = release.assets.Find(asset => asset.name.EndsWith(".zip"));
+                    if (zipAsset != null) {
+                        Match match = ServerReleaseVersionRegex().Match(release.body);
+                        if (match.Success) {
+                            string releasePatch = match.Groups[1].Value;
+                            release.tag_name = release.name + " - Tarkov Version: " + releasePatch;
```
Plootie, line 309: Can we replace this with string interpolation plz

Plootie (changes requested):
Good to go once the string interpolation is added

SIT.Manager.Avalonia/Services/InstallerService.cs
artehe: Yep done

Plootie (approved):

PR #15, opened 2024-02-24
Add update checks

Plootie:
I kinda hate how the bar looks tbh but idk how to change it

### DIFF
--- a/SIT.Manager.Avalonia/AssemblyInfo.cs
+++ b/SIT.Manager.Avalonia/AssemblyInfo.cs
@@ -1,0 +1,14 @@
+﻿using System.Reflection;
+using System.Runtime.Versioning;
+
+[assembly: AssemblyTitle("SIT.Manager.Avalonia")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("StayInTarkov")]
+[assembly: AssemblyProduct("SIT.Manager.Avalonia")]
+[assembly: AssemblyCopyright("Copyright © StayInTarkov 2024")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+[assembly: SupportedOSPlatform("windows")]
+[assembly: SupportedOSPlatform("linux")]
+[assembly: AssemblyVersion("0.0.0.0")]
--- a/SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
+++ b/SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
@@ -6,7 +6,6 @@ using SIT.Manager.Avalonia.Interfaces;
 using CommunityToolkit.Mvvm.Messaging;
 using FluentAvalonia.UI.Controls;
 using FluentAvalonia.UI.Media.Animation;
-using SIT.Manager.Avalonia.Interfaces;
 using SIT.Manager.Avalonia.Models;
 using SIT.Manager.Avalonia.Models.Messages;
 using System;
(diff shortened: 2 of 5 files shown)


SIT.Manager.Avalonia/Views/MainView.axaml
```diff
@@ -25,11 +25,14 @@
 			</ui:TeachingTip.IconSource>
 		</ui:TeachingTip>
 
-		<StackPanel Name="InfoBarStackPanel" Grid.Row="0" x:FieldModifier="Public" IsVisible="True">
-			<ui:InfoBar Name="UpdateInfoBar" IsOpen="False">
+		<StackPanel Name="InfoBarStackPanel" Grid.Row="0" x:FieldModifier="Public">
+			<ui:InfoBar Name="UpdateInfoBar" IsOpen="{Binding UpdateAvailable}" MaxHeight="45" Margin="0 0 0 10">
```
artehe, line 29: Feel like the the margin of 10 on the bottom of this bar makes it look the most weird. Reckon it should just be `Margin="0"` or could just have the whole Margin bit deleted.

SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
```diff
@@ -6,41 +6,68 @@
...
+        _logger = logger;
+        _httpClient = httpClient;
 
         _actionNotificationService.ActionNotificationReceived += ActionNotificationService_ActionNotificationReceived;
         _barNotificationService.BarNotificationReceived += BarNotificationService_BarNotificationReceived;
 
         WeakReferenceMessenger.Default.Register(this);
 
         UpdateButtonCommand = new AsyncRelayCommand(UpdateButton);
+        _ = CheckForUpdate();
```
artehe, line 54: Possibly more of a personal thing than anything, but I'm not the biggest fan of just calling the async task in the constructor like this. 

Two ways I think I'd be happier with this is we don't run the function on creation but when the view is actually activated similar to how we do things in the ModsPageViewModel.  Or the alternative would be to use something like this `RxApp.TaskpoolScheduler.Schedule(FetchReleases);`.

SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
```diff
@@ -6,41 +6,68 @@
...
+            string gitVersionString = await _httpClient.GetStringAsync(MANAGER_VERSION_URL);
+            Version gitVersion = new Version(gitVersionString);
+
+            UpdateAvailable = gitVersion.CompareTo(currentVersion) > 0;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "CheckForUpdate");
+        }
     }
```
artehe, line 71: This doesn't check to see if the check for updates setting is enabled or not.

artehe (changes requested):

SIT.Manager.Avalonia/AssemblyInfo.cs
```diff
@@ -0,0 +1,14 @@
...
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("StayInTarkov")]
+[assembly: AssemblyProduct("SIT.Manager.Avalonia")]
+[assembly: AssemblyCopyright("Copyright © StayInTarkov 2024")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+[assembly: SupportedOSPlatform("windows")]
+[assembly: SupportedOSPlatform("linux")]
+[assembly: AssemblyVersion("0.0.0.0")]
```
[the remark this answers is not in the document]
Plootie: I wasn't actually sure which project this needed to go in, but at least for the version this is the correct location

SIT.Manager.Avalonia/Views/MainView.axaml
Plootie: Thats my bad. I was trying to add a margin to correct the alignment of things a little but the margin didn't seem to do anything for it. The entire infobar is handled weirdly and i'm kinda fighting with it

SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
```diff
@@ -6,41 +6,87 @@
...
+        _logger = logger;
+        _httpClient = httpClient;
 
         _actionNotificationService.ActionNotificationReceived += ActionNotificationService_ActionNotificationReceived;
         _barNotificationService.BarNotificationReceived += BarNotificationService_BarNotificationReceived;
 
         WeakReferenceMessenger.Default.Register(this);
 
         UpdateButtonCommand = new AsyncRelayCommand(UpdateButton);
+        CloseButtonCommand = new AsyncRelayCommand(() => { UpdateAvailable = false; return Task.CompletedTask; });
```
artehe, line 65: This could just be a normal RelayCommand as we don't use any asynchronous methods in it.

SIT.Manager.Avalonia/Views/MainView.axaml
```diff
@@ -25,12 +25,18 @@
...
-					<Button Name="UpdateButton" Content="Update" Command="{Binding UpdateButtonCommand}" HorizontalAlignment="Right"/>
-				</ui:InfoBar.ActionButton>
-			</ui:InfoBar>
+		<StackPanel Name="InfoBarStackPanel" Grid.Row="0" x:FieldModifier="Public">
+			<Grid ColumnDefinitions="auto, auto, auto, *, auto" IsVisible="{Binding UpdateAvailable}">
+				<Canvas Margin="5 0 0 17" VerticalAlignment="Center" Grid.Column="0">
+					<Ellipse Fill="#1e4f8a" Width="18" Height="18"/>
+					<ui:SymbolIcon Symbol="Important" VerticalAlignment="Center" Margin="2 1 0 0"/>
+				</Canvas>
+				<TextBlock Text="An update to the SIT Manager is available click here to update:" Grid.Column="1" Margin="28 0 0 0" VerticalAlignment="Center"/>
```
artehe, line 34: Using a panel set with `Margin="14,0"` instead of a canvas here would avoid the need for having a margin set on the TextBlock, and also then line the icon up with the menu below and that would be consistent with how the bar notifications happen as well.

artehe (commented):
Almost there, just a few small tweaks to go 😄 

Something we may also want to do is add this in the app.xaml styles to make the default font for text that we display be consistent everywhere:


		<Style Selector="TextBlock">
			<Setter Property="FontFamily" Value="{StaticResource BenderFont}"/>
		</Style>

PR #23, opened 2024-02-26
Quick play changes

### LINKED ISSUE #3: Add some kind of "progress" indicator to the connect and quick play button
Crossposting https://github.com/Plootie/SIT.Manager.Avalonia/issues/30

Plootie:
Closes #3

### DIFF
--- a/SIT.Manager.Avalonia/Services/AkiServerService.cs
+++ b/SIT.Manager.Avalonia/Services/AkiServerService.cs
@@ -79,7 +79,11 @@ namespace SIT.Manager.Avalonia.Services
                 }
             });
             _process.OutputDataReceived += startedEventHandler;
-            _process.Exited += new EventHandler((sender, e) => ExitedEvent(sender, e));
+            _process.Exited += new EventHandler((sender, e) =>
+            {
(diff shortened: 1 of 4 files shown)


SIT.Manager.Avalonia/Services/AkiServerService.cs
```diff
@@ -79,15 +79,24 @@ public class AkiServerService(IBarNotificationService barNotificationService,
...
                 _ = Task.Run(() =>
                 {
+                    //This is gross but i genuinely have no other way to do this besides massively overcomplicated systems
                     System.Threading.Thread.Sleep(10 * 1000);
-                    ServerStarted?.Invoke(this, new EventArgs());
+                    if(!_process?.HasExited ?? true)
+                    {
+                        IsStarted = true;
+                        ServerStarted?.Invoke(this, new EventArgs());
+                    }
```
artehe, line 99: I have to agree with you but also can't think of a pleasant way of handling it currently. Might be an idea to add it as an issue to come back to in the future as I imagine if someones server has a few mods in then it potentially could take longer than 10 seconds to actually boot up.

artehe (approved):

PR #28, opened 2024-02-26
Server alive check

### LINKED ISSUE #27: Find better way of determining server launch status
I have to agree with you but also can't think of a pleasant way of handling it currently. Might be an idea to add it as an issue to come back to in the future as I imagine if someones server has a few mods in then it potentially could take longer than 10 seconds to actually boot up.

_Originally posted by @artehe in https://github.com/stayintarkov/SIT.Manager.Avalonia/pull/23#discussion_r1503093343_

artehe:
Improve on the progress made by @Plootie with the quick launch by providing a way of waiting for the server to be running rather than relying on a timer.

This should resolve #27 

Also sorry in advance I started moving the namepsaces of some files cause they felt wrong to me, then thought I should probably wait to move them. I can move them back if you disagree (I still think some more need moving but they can always happen later).


SIT.Manager.Avalonia/Classes/TarkovRequesting.cs
```diff
@@ -33,82 +29,87 @@ private async Task<Stream> Send(string url, HttpMethod? method = null, string? d
...
+            }
+
+            using MemoryStream ms = new();
+            await postStream.CopyToAsync(ms, cancellationToken);
+
+            string pingResponse = _compressionService.Decompress(ms.ToArray());
+            if (pingResponse.Equals("\"pong!\"")) {
+                return true;
+            }
+            return false;
```
Plootie, line 111: Nitpicky but could we directly return the equals check here?

SIT.Manager.Avalonia/Services/AkiServerService.cs
```diff
@@ -79,30 +87,51 @@ public class AkiServerService(IBarNotificationService barNotificationService,
...
-                    {
-                        IsStarted = true;
-                        ServerStarted?.Invoke(this, new EventArgs());
+            UpdateRunningState(RunningState.Starting);
+
+            if (cal) {
+                _ = Task.Run(async () => {
+                    TarkovRequesting requesting = ActivatorUtilities.CreateInstance<TarkovRequesting>(_serviceProvider, new Uri("http://127.0.0.1:6969/"));
+                    int retryCounter = 0;
+                    while (retryCounter < 6) {
```
Plootie, line 102: `retryCounter++`?

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -251,35 +230,39 @@ private async Task ConnectToServer(bool launchServer = false)
...
-                });
-
-                TimeSpan timeToAbort = default;
-                while ((timeToAbort = abortTime - DateTime.Now).TotalSeconds > 0 && !cts.IsCancellationRequested && !_akiServerService.IsStarted)
-                {
-                    QuickPlayText = $"Waiting for server ({timeToAbort.TotalSeconds:N0}s)";
-                    await Task.Delay(500);
+                bool aborted = false;
+                RunningState serverState = _akiServerService.State;
+                while (serverState == RunningState.Starting) {
```
Plootie, line 238: What do you think about moving the assignment to the condition? e.g `(serverState = _akiServerService.State) == RunningState.Starting`

Plootie (commented):
Just these 3 small nitpicks if you agree

SIT.Manager.Avalonia/Services/AkiServerService.cs
artehe: is at the bottom of the while loop, or is that not what you mean?

Plootie (approved):

PR #30, opened 2024-02-27
More sit manager porting

artehe:
This should be half of issue #26 porting over the bits which are relevant from the improvements PR.

Gonna need to test the SIT installation again and hope that nothing is now broken there again.


artehe:
That should be everything now. Although making it fullscreen does seem to overflow the edges a little bit like 8 pixels all round, I think I shall just make that an issue and come back to it.

Plootie (approved):
All good! Nice work, I didn't expect you to get this done so fast tbh

artehe:
Most of it we either did already, or had a template I could use to copy and paste the changes to.

PR #31, opened 2024-02-27
Little bit of refactoring, null check, window max size fix.

J0nathan550:
In this PR I was focused by shutting down all of the possible warnings. And removing dependencies that are useless in some classes. Still, there is more work need to be done. But at least I put a lot of _logger to not skip any of null exception.


Plootie (changes requested):
Could we restore the formatting? I can't tell what has been changed through the formatting changes

J0nathan550:
I will then just fix window then.

J0nathan550:
And return warnings, hang tight 😄

PR #33, opened 2024-02-28
Added not fully integrated localization logic, improved Page Navigation.

J0nathan550:
Fixed: Page Navigation that before was hard coded using string. Problem start to occure when I started developing localization for the Pane strings. I just noticed that when you clicked on buttons (in Pane) and they have different label in them it would not switch page because of the hardcoded Pane string content. It was fixed by adding events.

Added localization logic, the code itself works just fine with no problems. What is left is to bind everywhere the strings with localization we desire. I will continue working on the localization binding. But this is my safe push to keep up with other people. It should cost no problem in code whatsover.

Works by using DynamicResource binding, it grabs from Avalonia Resources specific file of language that you use (default is English) and translates it.

Pages translated: PlayView, MainView (Pane).

What to finish: Fully bind everywhere (where possible strings to translate), create big template English resource file to later add translation for other languages. Implement same localization binding but for the .cs source code (strings inside of source code are not translated, but I hope I will figure out soon how to translate them.).

Video demonstration of this PR:
https://youtu.be/XgD_EiMDU9Y


SIT.Manager.Avalonia/Models/LocalizationConfig.cs
```diff
@@ -0,0 +1,55 @@
...
+                    Translate("Ukrainian");
+                    break;
+                case Languages.Russian:
+                    Translate("Russian");
+                    break;
+                default:
+                    Translate("English");
+                    break;
+            }
+        }
```
Plootie, line 53: Models should contain minimal to no logic inside of them. This should be pushed into a service

SIT.Manager.Avalonia/Models/LocalizationConfig.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Avalonia.Markup.Xaml.Styling;
+using System.Linq;
+using System;
+
+namespace SIT.Manager.Avalonia.Models
+{
+    public static class LocalizationConfig
```
Plootie, line 7: I'm personally not a fan of marking the localization system as static and accessing it globally. This would be better added as a service and injected into classes that need it

SIT.Manager.Avalonia/Models/LocalizationConfig.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public enum Languages
```
Plootie, line 9: We should probably use CultureInfo over our own enums

SIT.Manager.Avalonia/ViewModels/MainViewModel.cs
```diff
@@ -50,6 +51,7 @@ public partial class MainViewModel : ViewModelBase, IRecipient<PageNavigationMes
         ILogger<MainViewModel> logger,
         HttpClient httpClient)
     {
+        UpdateTranslationStrings((Languages)managerConfigService.Config.CurrentLanguageSelected);
```
Plootie, line 54: Accessing the localization system like this gives me the ick, see previous comments about services

SIT.Manager.Avalonia/Views/MainView.axaml.cs
```diff
@@ -46,12 +32,53 @@ public partial class MainView : ReactiveUserControl<MainViewModel>
...
+    }
+
+    /// <summary>
+    /// <see cref="ModsButton_Clicked(object?, TappedEventArgs)"/> gets fired when you click on ModsButton in Pane View.
+    /// </summary>
+    private void ModsButton_Clicked(object? sender, TappedEventArgs e)
+    {
+        PageNavigation pageNavigation = new(typeof(ModsPage));
+        WeakReferenceMessenger.Default.Send(new PageNavigationMessage(pageNavigation));
+    }
```
Plootie, line 83: I don't see how this is any better than the current system. In the end it's still hard-coded types and events you're just moving where the hard coded types lie

SIT.Manager.Avalonia/Views/SettingsPage.axaml
```diff
@@ -31,6 +31,11 @@
 						<CheckBox Content="Look For Updates" 
 								  ToolTip.Tip="Automatically look for updates when opening the launcher."
 								  IsChecked="{Binding Config.LookForUpdates, Mode=TwoWay}"/>
+						<ComboBox Name="languageComboBox" ToolTip.Tip="Translates to your language, select one!" SelectedIndex="{Binding Config.CurrentLanguageSelected, Mode=TwoWay}" SelectionChanged="ComboBox_SelectionChanged">
+							<ComboBoxItem Content="English"/>
+							<ComboBoxItem Content="Ukrainian"/>
+							<ComboBoxItem Content="Russian"/>
```
Plootie, line 37: These item's should be loaded in from a list using data templating for maintainability

SIT.Manager.Avalonia/Views/SettingsPage.axaml
```diff
@@ -31,6 +31,11 @@
 						<CheckBox Content="Look For Updates" 
 								  ToolTip.Tip="Automatically look for updates when opening the launcher."
 								  IsChecked="{Binding Config.LookForUpdates, Mode=TwoWay}"/>
+						<ComboBox Name="languageComboBox" ToolTip.Tip="Translates to your language, select one!" SelectedIndex="{Binding Config.CurrentLanguageSelected, Mode=TwoWay}" SelectionChanged="ComboBox_SelectionChanged">
```
Plootie, line 34: I think we should stay away from using an event here. Use IPropertyChanged to notify when the SelectedItem is changed

SIT.Manager.Avalonia/Views/SettingsPage.axaml.cs
```diff
@@ -10,5 +11,10 @@ public partial class SettingsPage : UserControl
             InitializeComponent();
             this.DataContext = App.Current.Services.GetService<SettingsPageViewModel>();
         }
+
+        private void ComboBox_SelectionChanged(object? sender, SelectionChangedEventArgs e)
+        {
+            UpdateTranslationStrings((Languages)languageComboBox.SelectedIndex);
+        }
```
Plootie, line 18: See previous comment about avoid events

Plootie (changes requested):

SIT.Manager.Avalonia/Views/MainView.axaml.cs
J0nathan550: > I don't see how this is any better than the current system. In the end it's still hard-coded types and events you're just moving where the hard coded types lie

Better in they way that you can switch pages that are not depended literally on string. Current system checks what current page name is. And if the page name != whatever in Dictionary we cannot go to another page. Also why to bother? They are meant to be static. I don't think we will add 10.000+ buttons in that pane what so ever.

SIT.Manager.Avalonia/Models/LocalizationConfig.cs
J0nathan550: > We should probably use CultureInfo over our own enums

Problem behind Culture Info it's not 100% correct when you are starting the application. But still a good point to set at the very first start of the application the language based on what language you have on computer.

SIT.Manager.Avalonia/Views/SettingsPage.axaml.cs
J0nathan550: I just don't know how to use commands, this thing is complicated for me.

SIT.Manager.Avalonia/Views/MainView.axaml.cs
Plootie: This is actually my fault as I appear to have accidentally swapped from using the NavigationViewItem's Tag property to the InvokedItem which gives the display name. The Tag property allows us to have an identifier separate from the display value without needing an event for each item. I will push the needed changes for this now

Plootie:
I've made a [pull request](https://github.com/stayintarkov/SIT.Manager.Avalonia/pull/34) to fix the NavigationView problem. Once Artehe signs off on it I'll merge it into master for you to update from

J0nathan550:
> I've made a [pull request](https://github.com/stayintarkov/SIT.Manager.Avalonia/pull/34) to fix the NavigationView problem. Once Artehe signs off on it I'll merge it into master for you to update from

I saw that thank you, I will now avoid any events with this one 😄

J0nathan550:
I'm still working, I just fixed issue with my branch. No review required.

J0nathan550:
- Implemented MVVM, Service of Localization.
- Translated almost fully pages.
- Added just logic to be able Translate source code, still I have to translate entire source code. 
Hope you will like it.

Plootie (changes requested):
I'll hold off reviewing the code for now. This appears to be broken on my end. Selecting any language overwrites the option with "English (United States)" 
bNwEhVV4f7

J0nathan550:
I did one dumb mistake, sorry one second.

Plootie:
I'll review this in the morning, There's quite a few things I wanna comment on that'll take too long tonight

SIT.Manager.Avalonia/Views/MainWindow.axaml
```diff
@@ -13,8 +13,7 @@
 		Width="800"
 		Height="475"
 		WindowStartupLocation="CenterScreen"
-		TransparencyLevelHint="AcrylicBlur"
-		Background="Transparent"
+		Background="Black"
```
Plootie, line 16: Not sure why this is here. It's possible that these options aren't needed but I'm not a fan of changing things unrelated to the PR and undocumented

SIT.Manager.Avalonia/Views/SettingsPage.axaml
```diff
@@ -22,74 +23,88 @@
...
-						<CheckBox Content="Look For Updates" 
-								  ToolTip.Tip="Automatically look for updates when opening the launcher."
+						<CheckBox Content="{DynamicResource SettingsLookForUpdatesTitle}" 
+								  ToolTip.Tip="{DynamicResource SettingsLookForUpdatesTitleToolTip}"
 								  IsChecked="{Binding Config.LookForUpdates, Mode=TwoWay}"/>
+						<Grid>
+							<Grid.ColumnDefinitions>
+								<ColumnDefinition Width="auto"/>
+								<ColumnDefinition/>
+							</Grid.ColumnDefinitions>
```
Plootie, line 39: This can be condense into `<Grid ColumnDefinitions="auto, *">`

SIT.Manager.Avalonia/Views/PlayPage.axaml
```diff
@@ -6,21 +6,22 @@
...
-			<Button Name="ConnectButton" Command="{Binding ConnectToServerCommand}" Content="Connect" ToolTip.Tip="Connect to the server" Margin="10"/>
-			<Button Name="QuickPlayButton" Command="{Binding QuickPlayCommand}" ToolTip.Tip="Start the Aki server and connect to the server" Margin="10">
+			<Label Content="{DynamicResource PasswordBoxTitle}" Margin="10 10 10 0" FontFamily="{StaticResource BenderFont}" FontStyle="Oblique"/>
+			<TextBox Name="PasswordBox" Text="{Binding Password, Mode=TwoWay}" Margin="10 5 10 10" HorizontalAlignment="Left" PasswordChar="●" Watermark="{DynamicResource EnterPasswordPlaceholder}"/>
+			<CheckBox Name="RememberMeCheck" Content="{DynamicResource RememberMeCheckBoxTitle}" Margin="10" IsChecked="{Binding RememberMe}"/>
+			<Button Name="ConnectButton" Command="{Binding ConnectToServerCommand}" Content="{DynamicResource ConnectButtonTitle}" ToolTip.Tip="{DynamicResource ConnectButtonTitleToolTip}" Margin="10"/>
+			<Button Name="QuickPlayButton" Command="{Binding QuickPlayCommand}" ToolTip.Tip="{DynamicResource StartServerAndConnectButtonTitleToolTip}" Margin="10">
```
Plootie, line 24: Removal of the width attributes here causes the textboxs to shrink to minimum size when you enter a single character as its resizing to largest content and once a character is entered the watermark disappears causing the shrink

SIT.Manager.Avalonia/Models/LocalizationModel.cs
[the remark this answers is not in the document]
J0nathan550: But you cannot store everything in CultureInfo, because it's gonna be a waste of memory, better to store few variables. Also it's gonna be for me at least a lot harder to understand and save literally CultureInfo.

SIT.Manager.Avalonia/Views/MainWindow.axaml
J0nathan550: I think it's better to remove AcrylicBlur because we literally do not use transparent AcrylicBlur background, it can make a difference in perfomance when you render a window.

SIT.Manager.Avalonia/Views/SettingsPage.axaml
J0nathan550: Sugar syntax but okay 😄

SIT.Manager.Avalonia/Models/LocalizationModel.cs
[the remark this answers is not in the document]
J0nathan550: But I will see, maybe it's not that hard.

SIT.Manager.Avalonia/Views/MainWindow.axaml
Plootie: Can we move this to another PR. I don't disagree that these are possibly useless but it's outside the scope of this PR and I don't like making undocumented changes

SIT.Manager.Avalonia/Models/ManagerConfig.cs
```diff
@@ -28,7 +28,7 @@ public partial class ManagerConfig : ObservableObject
         [ObservableProperty]
         public bool _lookForUpdates = true;
         [ObservableProperty]
-        public LocalizationModel _currentLanguageSelected = new() { FullNameLanguage = CultureInfo.CurrentCulture.NativeName, ShortNameLanguage = CultureInfo.CurrentCulture.Name };
+        public string _currentLanguageSelected = CultureInfo.CurrentCulture.Name;
```
Plootie, line 31: This falls back to the users current culture info, but if the user's culture doesn't match any of the localizations we have the selected field will be blank (In my case this defaults to en-GB, since that doesn't match any localizations we have the selected item becomes blank)

SIT.Manager.Avalonia/Services/LocalizationService.cs
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    result = (string)translation;
+                    for (int i = 0; i < replaces.Length; i++)
+                    {
+                        result = result.Replace($"%{i + 1}", replaces[i]);
+                    }
+                }
+            }
+            return result;
+        }
```
Plootie, line 74: These two methods have significant overlap, Push some of the duplicate code to another function to avoid repeating the same thing. Also instead of using a try catch were reference the known available localizations to ensure it's valid before attempting

SIT.Manager.Avalonia/Services/LocalizationService.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// </summary>
+        /// <returns>Resource with Localization</returns>
+        private ResourceInclude CreateResourceLocalization(string url)
+        {
+            var self = new Uri("resm:Styles?assembly=SIT.Manager.Avalonia");
+            return new ResourceInclude(self)
+            {
+                Source = new Uri(url)
+            };
+        }
```
Plootie, line 117: I don't like this approach. We're repeating ourselves massively here. We already pull values such as `SIT.Manager.Avalonia.Localization.en-US.axaml` from the first half of the LINQ expression, only to extract the country code, convert it into a CultureInfo to the reconstruct that exact path again later to create the resource

SIT.Manager.Avalonia/ViewModels/SettingsPageViewModel.cs
```diff
@@ -107,7 +122,8 @@ public partial class SettingsPageViewModel : ViewModelBase
 
     protected override void OnPropertyChanged(PropertyChangedEventArgs e) {
         base.OnPropertyChanged(e);
-
+        _localizationService.Translate(CurrentLocalization);
```
Plootie, line 125: This will call the localization code on every property change regardless of if the localization was actually changed

Plootie (changes requested):

SIT.Manager.Avalonia/Models/ManagerConfig.cs
J0nathan550: Yeah, I forgot about that issue will fix it.

SIT.Manager.Avalonia/ViewModels/SettingsPageViewModel.cs
J0nathan550: Sure, sorry, I will add propertychange based on if the Settings current localization was changed.

SIT.Manager.Avalonia/Services/LocalizationService.cs
Plootie, line 1: My only gripe left with this PR is some of these methods. I'm seeing a lot of `avares://SIT.Manager.Avalonia/Localization/*.axaml` that eventually ends up getting passed down to `CreateResourceLocalization`. This could be changed so that only the file name gets passed in to the method and it constructs the rest of the path itself. Once that's addressed I'm happy to merge this ^^

Plootie (changes requested):

SIT.Manager.Avalonia/Services/LocalizationService.cs
J0nathan550: Do you want me to later recreate PR to avoid thousands of commits?

J0nathan550:
Done!

Plootie (approved):
I think I'm happy with this :3 @artehe Do you mind just giving this a skim to make sure I haven't missed something dumb?

SIT.Manager.Avalonia/ViewModels/SettingsPageViewModel.cs
```diff
@@ -107,7 +122,11 @@ public partial class SettingsPageViewModel : ViewModelBase
 
     protected override void OnPropertyChanged(PropertyChangedEventArgs e) {
         base.OnPropertyChanged(e);
-
+        if (e.PropertyName == nameof(CurrentLocalization))
+        {
+            _localizationService.Translate(CurrentLocalization);
+            Config.CurrentLanguageSelected = CurrentLocalization.Name;
+        }
```
artehe, line 129: I know I said two things but I can't think how to handle the second so we can leave it for now and I'll create an issue to come back to it sometime in the future.  So one last thing then this is good to go. 

Could we take the highlighted lines out and instead use th the specific property change value for CurrentLanguageSelected?

If you wanna just copy and paste the function it'd be this:
```
    partial void OnCurrentLocalizationChanged(CultureInfo value)
    {
        _localizationService.Translate(value);
        Config.CurrentLanguageSelected = value.Name;
    }
```

Plootie (approved):

PR #34, opened 2024-02-28
Route pages based on NavViewItem's tag

Plootie:
This was the intended method that I appear to have mixed up with the InvokedItem at some point. Should allow for localization


Plootie:
@artehe Can you just double check this won't break anything massively before I merge it?

artehe:
> @artehe Can you just double check this won't break anything massively before I merge it?

You're good, that should work fine.

PR #38, opened 2024-02-28
Overhaul TarkovRequesting + PlayPage

Plootie:
Due to @artehe finding a bunch of exposed endpoints for the server I'll be rewriting TarkovRequesting to play nicely with the usage of theses new requests and rebuild the playpage to facilitate new features


SIT.Manager/Models/Aki/AkiCharacter.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+#pragma warning disable CA1507 // Use nameof to express symbol names
+    [JsonProperty("Username")]
+    public string Username { get; init; } = username;
+    [JsonProperty("Password")]
+    public string Password { get; init; } = password;
+    [JsonProperty("ProfileID")]
+    public string ProfileID { get; internal set; } = "";
+    [JsonProperty("Edition")]
+#pragma warning restore CA1507 // Use nameof to express symbol names
```
artehe, line 20: Do these JsonProperty names need the capitalization? As the other models don't use them see SIT.Manager/Models/Aki/AkiMiniProfile.cs or in the case if SIT.Manager/Models/Aki/AkiServer.cs you could just use nameof.

Either way would be nicer to not have to use ```#pragma warning disable CA1507```

SIT.Manager/Models/Aki/AkiServer.cs
```diff
@@ -0,0 +1,20 @@
+﻿using CommunityToolkit.Mvvm.ComponentModel;
+using Newtonsoft.Json;
```
artehe, line 2: Something I just noticed as well is you're using Newtonsoft.Json I think in a lot of other places we use System.Text.Json so maybe we wanna make sure we pick one or the other for the whole thing rather than using both

SIT.Manager/SIT.Manager.csproj
```diff
@@ -19,6 +19,10 @@
 		<AvaloniaResource Include="Assets\**" />
 		<Compile Include="..\SIT.Manager.Desktop\AssemblyInfo.cs" Link="AssemblyInfo.cs" />
 		<EmbeddedResource Include="Localization\*.axaml" />
+		<AvaloniaXaml Remove="Classes\**" />
+		<Compile Remove="Classes\**" />
+		<EmbeddedResource Remove="Classes\**" />
+		<None Remove="Classes\**" />
```
artehe, line 25: These lines can be removed, its a little pet peeve of mine is how happy the csproj files are to add stuff which is completely useless and makes them so messy.

SIT.Manager/Services/AkiServerRequestingService.cs
```diff
@@ -0,0 +1,162 @@
...
+    private async Task<string> LoginOrRegisterAsync(AkiCharacter character, string operation, CancellationToken cancellationToken = default)
+    {
+        using (MemoryStream ms = await SendAsync(character.ParentServer, Path.Combine("/launcher/profile", operation), HttpMethod.Post, CreateLoginData(character), cancellationToken: cancellationToken))
+        using (StreamReader streamReader = new StreamReader(ms))
+        {
+            string response = await streamReader.ReadToEndAsync(cancellationToken);
+            return response.ToLowerInvariant() switch
+            {
+                "invalid_password" => throw new IncorrectAccountPasswordException(), //Should these really be exceptions?...
+                "failed" => throw new AccountNotFoundException(),
```
artehe, line 134: As your comment suggests they don't necessarily need to be exceptions you could probably achieve the same result by creating an enum with possible outputs and then return an object which contains this enum result and any other info that needs returning then rather than having to handle an exception you just parse the result data.

Which method is better I don't know.

artehe (commented):
Just a few comments at the moment as it's still a work in progress, mainly just some things to think about and some small issues.

SIT.Manager/Models/Aki/AkiServer.cs
Plootie: I kinda just default to using newtonsoft as I've always used it but i dunno if we *should* be using it. I'd need your take on which you'd prefer to use

SIT.Manager/Models/Aki/AkiCharacter.cs
Plootie: AkiMiniProfile uses lowercase because it needs to deserialize a server response. AkiServer idk why it went with nameof here honestly, My main concern with using nameof is that if we ever change the name of the variable then anything serialized with the previous variable name will now be incompatible which doesn't seem like a good design choice

SIT.Manager/App.axaml.cs
```diff
@@ -30,84 +34,133 @@ public sealed partial class App : Application
...
+
+            #region Services
+
+            services.AddSingleton<IActionNotificationService, ActionNotificationService>()
+                .AddSingleton<IAkiServerService, AkiServerService>()
+                .AddSingleton<ITarkovClientService, TarkovClientService>()
+                .AddSingleton<IBarNotificationService, BarNotificationService>()
+                .AddTransient<IFileService, FileService>()
+                .AddTransient<IInstallerService, InstallerService>()
+                .AddSingleton<IManagerConfigService, ManagerConfigService>()
```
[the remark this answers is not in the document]
Plootie: Ty. I must have missed this while merging the like 200 commits from master @.@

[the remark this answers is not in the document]
artehe: Happens 😆

SIT.Manager/Models/Aki/AkiCharacter.cs
artehe: That's a fair point probably makes sense to not use nameof for these then

SIT.Manager/Models/Aki/AkiServer.cs
artehe: A couple of years ago 100% newtonsoft would have been my answer.

I think now they're basically the same in terms of how you use them and I believe System.Text.Json is slightly more performant as well. Plus it's also baked in rather than having to be a nuget package (not that that is an actual positive or negative)

I tend to just prefer System.Text.Json now but don't think there is much in it.

Plootie: Cool 👍 I'll remove the newtonsoft package and see if swapping to the native lib breaks anything

PR #39, opened 2024-02-28
Preserve server log

### LINKED ISSUE #22: Server log isn't preserved when launched from quick play
If the server is launched via the quick play button and the user never visits the server page the server output won't be logged until the page is visited

artehe:
Should resolve #22 

I think this is the nicest way of doing it; I couldn't come up with any other way of handling it at least.


Plootie:
Once I finish eating I'll test the changes but everything on the code side looks good

artehe:
Should fix the manager not waiting for server and some other things that broke as a result of me fiddling with these bits

Plootie (approved):

PR #41, opened 2024-02-29
Fixes for linux #1

artehe:
There's almost certainly more but these were two things I found on Linux yesterday.

The first is that the patcher could be called patcher.exe rather than Patcher.exe.

Then the second is if the zip file was compressed using lzma it would fail but with the SharpCompress package it's fine. The original manager used this but somehow I missed that.

### DIFF
--- a/SIT.Manager.Avalonia/Services/InstallerService.cs
+++ b/SIT.Manager.Avalonia/Services/InstallerService.cs
@@ -100,14 +100,18 @@ namespace SIT.Manager.Avalonia.Services
         /// Runs the downgrade patcher
         /// </summary>
         /// <returns>string with result</returns>
-        private async Task<string> RunPatcher() {
+        private async Task<string> RunPatcher()
+        {
             _logger.LogInformation("Starting Patcher");
             _actionNotificationService.StartActionNotification();
             _actionNotificationService.UpdateActionNotification(new ActionNotification("Running Patcher...", 100));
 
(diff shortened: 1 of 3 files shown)


SIT.Manager.Avalonia/Services/InstallerService.cs
```diff
@@ -100,15 +100,19 @@ public partial class InstallerService(IActionNotificationService actionNotificat
...
             _logger.LogInformation("Starting Patcher");
             _actionNotificationService.StartActionNotification();
             _actionNotificationService.UpdateActionNotification(new ActionNotification("Running Patcher...", 100));
 
             string patcherPath = Path.Combine(_configService.Config.InstallPath, "Patcher.exe");
             if (!File.Exists(patcherPath)) {
+                patcherPath = Path.Combine(_configService.Config.InstallPath, "patcher.exe");
+                if (!File.Exists(patcherPath)) {
                 return $"Patcher.exe not found at {patcherPath}";
             }
```
Plootie, line 114: It took a lot of doc reading but there is a `Directory.GetFiles` overload that supports an `EnumerationOptions` argument which has the `MatchCasing` property you can set to false which might make this a little neater. [MS Docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=netcore-3.1#system-io-directory-getfiles(system-string-system-string-system-io-enumerationoptions))

artehe: Good find!

Changed it to work with that now

PR #49, opened 2024-03-01
Fix server detecting

Plootie:
Remove old irrelevant detection code


Plootie:
@J0nathan550 Can you give this a try and see if it fixes it? I don't know if this actually is the issue but it's my first assumption and needed removing anyway

J0nathan550:
No doesn't fix issue.

J0nathan550 (changes requested):
Works, but also add ability in ServerViewModelPage ability to cancel while it's initializing the server. Please

J0nathan550 (approved):
Works now as it should 😄

PR #54, opened 2024-03-02
Updated uk-UA, ru-RU missing keys.

J0nathan550:
literally forgot to update also Linux localization settings for those languages. Sorry...


Plootie:
Will @JimWails need up update the Chinese localizations?

JimWails:
> Will @JimWails need up update the Chinese localizations?

Already contain these keys

PR #56, opened 2024-03-02
Changed Titlebar Color of Window.

J0nathan550:
Changed Titlebar color of the Window to pure black. Better then white bright color that doesn't match the style of the launcher.

Works both on Windows and Linux.
Linux Mint  Работает  - Oracle VM VirtualBox 02 03 2024 18_54_42
SIT Manager 02 03 2024 18_51_32


SIT.Manager.Avalonia/Views/MainWindow.axaml.cs
```diff
@@ -1,14 +1,27 @@
...
+        TitleBar.InactiveBackgroundColor = new Color(0xFF, 0x00, 0x00, 0x00);
+        TitleBar.InactiveForegroundColor = new Color(0xFF, 0xFF, 0xFF, 0xFF);
+        TitleBar.ButtonBackgroundColor = new Color(0xFF, 0x00, 0x00, 0x00);
+        TitleBar.ButtonForegroundColor = new Color(0xFF, 0xFF, 0xFF, 0xFF);
+        TitleBar.ButtonHoverBackgroundColor = new Color(0xFF, 0x11, 0x11, 0x11);
+        TitleBar.ButtonHoverForegroundColor = new Color(0xFF, 0xFF, 0xFF, 0xFF);
+        TitleBar.ButtonPressedBackgroundColor = new Color(0xFF, 0x21, 0x21, 0x21);
+        TitleBar.ButtonPressedForegroundColor = new Color(0xFF, 0xFF, 0xFF, 0xFF);
+        TitleBar.ButtonInactiveBackgroundColor = new Color(0xFF, 0x00, 0x00, 0x00);
+        TitleBar.ButtonInactiveForegroundColor = new Color(0xFF, 0xFF, 0xFF, 0xFF);
```
Plootie, line 24: Are these exposed in the xaml at all? I'd rather use xaml for this if possible

J0nathan550: Let me see...

J0nathan550: It is, but it's not recognizable by the intellisense. Let me try to fix that.

J0nathan550: No, just tested it doesn't want to grab those values from xaml. :( we have to use it only through source.

J0nathan550:
Maybe, I am blind, here is the documentation.
https://amwx.github.io/FluentAvaloniaDocs/pages/Controls/AppWindow#related-enumsclasses

SIT.Manager.Avalonia/Views/MainWindow.axaml.cs
J0nathan550: Maybe, I am blind, here is the documentation.
https://amwx.github.io/FluentAvaloniaDocs/pages/Controls/AppWindow#related-enumsclasses

J0nathan550: No chance, by the way UWP has done it same way of accessing color elements. You cannot change them in xaml file, but you only can change them in initialization of the source code. So bad luck on this one.

J0nathan550:
@Plootie, there no way to access those items through xaml. Do you wish to create just function with name "Create Titlebar" or else?

artehe:
> @Plootie, there no way to access those items through xaml. Do you wish to create just function with name "Create Titlebar" or else?

Ye just putting them into a function called CreateTitlebar would work for me, and I think Plootie would agree.

SIT.Manager.Avalonia/App.axaml
```diff
@@ -23,7 +24,7 @@
 	</Application.Resources>
 	
     <Application.Styles>
-		<sty:FluentAvaloniaTheme />
+		<sty:FluentAvaloniaTheme CustomAccentColor="#323232" />
```
artehe, line 27: Do we really want to change this..... Makes some things really hard to see..... I'm not against the color being changed just not convinced a dark grey is the right choice.

With this change
image

Without:
image



And probably a more obvious example of it being hard to see:
With:
image

Without:
image

J0nathan550:
I don't like purple for launcher that uses black-white theme. Better to make the accent color white. We will able to see everything.

J0nathan550:
> I don't like purple for launcher that uses black-white theme. Better to make the accent color white. We will able to see everything.

SIT Manager 04 03 2024 14_10_48

Plootie:
Finally got myself a change to properly look over this now that the linux stuff is resolved (mostly). Despite all my effort I believe you are correct that there is absolutely no way to access those properties on the xaml side which is truly horrifying to me. I did some experimenting and found `#7f7f7f` seemed to be a good intermediate between the two for now. It would be interesting if we could allow users to edit this themselves through the settings menu at some point

J0nathan550:
> 7f7f7f

this color is awesome let me push again

PR #60, opened 2024-03-04
use the color picker provided by FluentAvalonia

### LINKED ISSUE #25: Solve binding error
I've noticed this for awhile but ignored it as it seems to not affect anything we're doing but I'd really like to find and solve this for the sake of not having rouge errors in the project. At runtime the error `[Binding]Error in binding to 'Avalonia.Controls.ColumnDefinition'.'Width': 'Null value in expression '{empty}' at '$self.TemplatedParent'.'(ColumnDefinition #7689721)` is thrown. I believe this is from the color picker

artehe:
Should resolve #25 by using the inbuilt color picker that came with FluentAvalonia. 

Haven't looked too closely at any binding errors, but this does let me remove a lot of xaml so that's nice.


Plootie:
Bizarrely this actually seems to create a secondary binding error. I'm all for pushing this anyway as these errors seemingly don't actually affect anything but it does really bug me that we don't know where they're coming from

Plootie (approved):

PR #63, opened 2024-03-04
Add manager crash protection for linux launching

Plootie:
This should hopefully protect against mistyped paths and the such


artehe (approved):
Probably the nicest way to handle it.

PR #66, opened 2024-03-04
Add warning when no downgrader is found

Plootie:


SIT.Manager.Avalonia/Localization/en-US.axaml
```diff
@@ -303,6 +303,7 @@
 	<system:String x:Key="InstallServiceInstallSuccessfulTitle">Install</system:String>
 	<system:String x:Key="InstallServiceInstallSuccessfulDescription">Installation of Server was successful.</system:String>
 	<system:String x:Key="InstallServiceInstallSITSuccessfulDescription">Installation of SIT was successful.</system:String>
+	<system:String x:Key="InstallerServiceNoDowngradePatcher">No downgrader is available for these versions of Tarkov.</system:String>
```
artehe, line 306: Would it not be more accurate to say "No downgrader is available for this version of Tarkov."?

Plootie: I was thinking about the fact that technically the downpatcher depends on both the current and target version

artehe (approved):

PR #69, opened 2024-03-04
Center button contents

Plootie:
Would love both of your input on this @artehe @J0nathan550 I can't tell if centered looks better or not

### DIFF
--- a/SIT.Manager.Avalonia/Views/ToolsPage.axaml
+++ b/SIT.Manager.Avalonia/Views/ToolsPage.axaml
@@ -32,7 +32,7 @@
 					Classes="toolsButton"
 					ToolTip.Tip="{DynamicResource ToolsSITToolTip}"
 					Command="{Binding InstallSITCommand}">
-				<StackPanel Orientation="Horizontal">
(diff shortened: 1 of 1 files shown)


J0nathan550:
If icons would be on same position on left and text on center this would look good! 😄

Plootie:
I did try that initially but since they're inside a stack panel a center align wont work so it'd require some kind of change to the buttons

J0nathan550:
> I did try that initially but since they're inside a stack panel a center align wont work so it'd require some kind of change to the buttons

You can use grid :D

Plootie:
Idk if my brain is just cooked tonight but using a grid set to `auto, *` wouldn't center align the text since the content align would be to the column it's in and I can't think of a better way. I'm sure there is one I just can't think of how to do this tonight

Plootie:
I've returned to this and determined with the size constraints there probably isn't a good way to have the icons be left aligned and text center aligned (`size, * , size` would technically work but wouldn't work for the larger texts). Do you think this should be merged or closed?

artehe:
I'd say just close it for now, gonna have some tweaks to that page soon ish anyway I think if I get round to improving the install ui stuffs

PR #70, opened 2024-03-04
Fix config wiping saved credentials

Plootie:
I was wondering why my credentials kept resetting and it's because I'm stupid. If we didn't specific to remember login when updating the config it would default to false instead of the last used value, meaning any config not explicitly stating to remember would null out saved credentials


artehe (approved):
Ah yes missed this before.

I do wonder if there's still some other things going on if we have many updates in quick succession

Issue #65, opened 2024-03-04
Implement ability to change accent color in application.


J0nathan550:
I will later myself (when I will have a free time) implement the color picker that will allow users to change accent color of the application.

PR #72, opened 2024-03-05
Added ability to change color of accent, fixed installer service dialog

J0nathan550:
Fixes issue: #65 
Allows to change the accent of selected elements in FluentAvalonia.
The item is translated on all languages.

accent


SIT.Manager.Avalonia/ViewModels/SettingsPageViewModel.cs
```diff
@@ -133,4 +141,14 @@ public partial class SettingsPageViewModel : ViewModelBase
         _localizationService.Translate(value);
         Config.CurrentLanguageSelected = value.Name;
     }
+
+    partial void OnCurrentAccentColorChanged(Color? value)
+    {
+        if (faTheme == null) faTheme = Application.Current?.Styles.OfType<FluentAvaloniaTheme>().FirstOrDefault();
```
Plootie, line 147: I've never observed this trigger, Is this needed?

Plootie (changes requested):
Only request is that you change the binding to `Config.CurrentAccentColor` and apply the accent inside the config changed event similar to how the console font color is handled

SIT.Manager.Avalonia/ViewModels/SettingsPageViewModel.cs
J0nathan550: > I've never observed this trigger, Is this needed?

For the theme specificly yes, I dig for one hour trying to get the FluentAvaloniaTheme. Lucky me, I found it.

J0nathan550:
Refactored. Also fixed one dialog in Installer Service, I was trying to play game with my brother, but noticed that when he had wrong version of Tarkov it pop up him a dialog that would not give him brief dialog that the current version of tarkov is downgraded, plus there was no button at all to close the dialog which locked application forever. Translated it, and fix it.

Plootie (approved):

Issue #67, opened 2024-03-04
Implement ability to hide ip address if person really needs it


J0nathan550:
Feature that was added in original SIT.Manager, that I myself implemented is ability through settings hide ip address if user wants it.
Hiding of ip will be added by default, but if user don't care about privacy he can toggle it off. (Will also implement myself later on)

PR #73, opened 2024-03-06
Implemented ability to hide ip or not.

J0nathan550:
Solves the issue: #67

Demonstration:
hideip

### DIFF
--- a/SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
+++ b/SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
@@ -1,4 +1,5 @@
-﻿using Avalonia.Controls.ApplicationLifetimes;
+﻿using Avalonia.Controls;
+using Avalonia.Controls.ApplicationLifetimes;
 using Avalonia.Threading;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
@@ -27,8 +28,8 @@ namespace SIT.Manager.Avalonia.ViewModels
         //TODO: Merge these constants into one play. Pretty sure we delcare at least one of these somewhere else already
         private const string SIT_DLL_FILENAME = "StayInTarkov.dll";
         private const string EFT_EXE_FILENAME = "EscapeFromTarkov.exe";
-        private readonly IManagerConfigService _configService;
         private readonly IServiceProvider _serviceProvider;
+        public readonly IManagerConfigService _configService;
 
         [ObservableProperty]
         private string _lastServer;
@@ -51,6 +52,7 @@ namespace SIT.Manager.Avalonia.ViewModels
         private readonly IAkiServerService _akiServerService;
         private readonly ILocalizationService _localizationService;
         private readonly ILogger<PlayPageViewModel> _logger;
+        private TextBox? addressBox;
 
         public IAsyncRelayCommand ConnectToServerCommand { get; }
         public IAsyncRelayCommand QuickPlayCommand { get; }
@@ -63,7 +65,7 @@ namespace SIT.Manager.Avalonia.ViewModels
             IAkiServerService akiServerService,
             ILocalizationService localizationService,
             ILogger<PlayPageViewModel> logger,
-            IServiceProvider serviceProvider) 
+            IServiceProvider serviceProvider)
         {
             _configService = configService;
             //TODO: Check that this is the best way to implement DI for the TarkovRequesting. Prettysure service provider would be better
@@ -79,6 +81,7 @@ namespace SIT.Manager.Avalonia.ViewModels
             _configService.ConfigChanged += (o, e) =>
             {
                 QuickPlayText = _localizationService.TranslateSource("PlayPageViewModelQuickPlayText");
+                if (addressBox != null) addressBox.RevealPassword = !_configService.Config.HideIpAddress;
             };
 
             _lastServer = _configService.Config.LastServer;
@@ -105,7 +108,8 @@ namespace SIT.Manager.Avalonia.ViewModels
             };
 
             string launchArguments = string.Join(' ', argumentList.Select(argument => $"{argument.Key}={argument.Value}"));
-            if (OperatingSystem.IsLinux()) {
+            if (OperatingSystem.IsLinux())
+            {
                 // We need to make sure that the json is contained in quotes on Linux otherwise you won't be able to connect to the server.
                 launchArguments = string.Join(' ', argumentList.Select(argument => $"{argument.Key}=\"{argument.Value}\""));
             }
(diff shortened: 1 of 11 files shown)


Plootie (changes requested):
Could you bind the reveal password property in xaml to an observable field instead?

J0nathan550:
> Could you bind the reveal password property in xaml to an observable field instead?



> Could you bind the reveal password property in xaml to an observable field instead?

No, that would be problematic, because I will lose the ability to have that reveal effect. If you know the way to make it observable as you want show me the code, I can't came up with the any single solution but this one sorry :(

J0nathan550:
Maybe this one is better?

Plootie:
Might be incorrect in saying this so maybe @artehe could lmk if this is unnecessary but I think a multibinding would work here instead of using the events

J0nathan550:
But still, I don't really understand what is wrong with this PR. I can't find better solution to this. You need settings to store if you wish to Hide IP or not. Which with last PR we do that, if want to hide IP we enable in event blinking effect of revealing IP, if we are not hiding, we just always reveal password. Don't see issue in here.

J0nathan550:
@Plootie

Plootie:
I don't like this this PR's reliance on the event subscriptions. There are multibinding's that solve this exact problem

J0nathan550:
I would be appreciated to learn about. This is a new thing to me. Sounds complicated...

Plootie:
I've never needed to use them myself, I just know they exist and vaguely how to use them. Some examples [here](https://github.com/AvaloniaUI/Avalonia/blob/cc714a203c3a16310bd4104a38f8a525588f65a2/src/Avalonia.Themes.Default/TextBox.xaml#L25) and [here](https://github.com/AvaloniaUI/Avalonia/discussions/13045)

J0nathan550:
@artehe please save me :D I don't know at all how to do multibindings.

artehe:
> @artehe please save me :D I don't know at all how to do multibindings.

They're pretty cool, and I always forget they exist 😓 

Probably what you want is something like this:
```xaml

<TextBox Name="AddressBox" Text="{Binding LastServer, Mode=TwoWay}" MinWidth="300" Margin="10 5 10 10" HorizontalAlignment="Left" Watermark="{DynamicResource EnterServerAddressPlaceholder}" PasswordChar="*">
    <TextBox.RevealPassword>
        <MultiBinding Converter="{x:Static BoolConverters.Or}">
	    <Binding RelativeSource="{RelativeSource TemplatedParent}" Path="IsFocused" />
	    <Binding Path="!ManagerConfig.HideIpAddress"/>
        </MultiBinding>
    </TextBox.RevealPassword>
</TextBox>
```

Instead of 
```xaml
<TextBox Name="AddressBox" Text="{Binding LastServer, Mode=TwoWay}" MinWidth="300" Margin="10 5 10 10" HorizontalAlignment="Left" Watermark="{DynamicResource EnterServerAddressPlaceholder}" PasswordChar="*" RevealPassword="{Binding !ManagerConfig.HideIpAddress}"/>
```

Then you can get rid of the events in PlayPage.axaml.cs and also set _configService back to private in PlayPageViewModel

Plootie:
> Collaborator

I don't know if what I did is _strictly correct_ but I got a compile error about the `TemplatedParent` relative source so I swapped that binding out for `<Binding Path="$parent.IsFocused" />` which seems to work. If there's a better lmk :3

artehe:
Oh ye that's probably what you do actually want my bad. I just typed that out without actually testing it first 😆

J0nathan550:
Well I'm fixing now the code :D

J0nathan550:
It gives me the enourmous behavior at first you start it works as intended but when you don't want to hide IP address and you select textbox and deselect it, it will hide the ip. And if you go back to it again the textbox will be hidden always. I still think there is literally no way to do it without any events.

J0nathan550:
I just literally need the logic that if the HideIpAddress is turned on it will depend on the IsFocus paramater, if the HideIpAddress is false it will literally just always show the IpAddress. So complicated with those multibindings. :(

Plootie:
> I just literally need the logic that if the HideIpAddress is turned on it will depend on the IsFocus paramater, if the HideIpAddress is false it will literally just always show the IpAddress. So complicated with those multibindings. :(

```
<TextBox Name="AddressBox" Text="{Binding LastServer, Mode=TwoWay}" MinWidth="300" Margin="10 5 10 10" HorizontalAlignment="Left" Watermark="{DynamicResource EnterServerAddressPlaceholder}" PasswordChar="*">
	<TextBox.RevealPassword>
		<MultiBinding Converter="{x:Static BoolConverters.Or}">
			<Binding Path="#AddressBox.IsFocused"/>
			<Binding Path="!ManagerConfig.HideIpAddress"/>
		</MultiBinding>
	</TextBox.RevealPassword>
</TextBox>
```
👍

J0nathan550:
> <TextBox Name="AddressBox" Text="{Binding LastServer, Mode=TwoWay}" MinWidth="300" Margin="10 5 10 10" HorizontalAlignment="Left" Watermark="{DynamicResource EnterServerAddressPlaceholder}" PasswordChar="*">
> 	<TextBox.RevealPassword>
> 		<MultiBinding Converter="{x:Static BoolConverters.Or}">
> 			<Binding Path="#AddressBox.IsFocused"/>
> 			<Binding Path="!ManagerConfig.HideIpAddress"/>
> 		</MultiBinding>
> 	</TextBox.RevealPassword>
> </TextBox>

Little bit better, but still it doesn't fix the issue when the HideIpAddress is turned off here is the demo:

https://github.com/stayintarkov/SIT.Manager.Avalonia/assets/88881607/2b41cfa9-ff4e-411b-b2b1-fcf5d73200c7

Plootie:
Possible Avalonia bug? 🤷‍♀️

J0nathan550:
> Possible Avalonia bug? 🤷‍♀️

Don't know, but anyway we need at least lost focus event to prevent that Avalonia bug. 🤣

Plootie:
I'd make an issue on avalonia to double check this is in fact an issue and if it turns out to be an avalonia bug we'll have to revert to events until it's fixed

J0nathan550:
> I'd make an issue on avalonia to double check this is in fact an issue and if it turns out to be an avalonia bug we'll have to revert to events until it's fixed

Okay :)

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -43,8 +44,14 @@ public partial class PlayPageViewModel : ViewModelBase
...
         private bool _rememberMe;
 
+        [ObservableProperty]
+        private ManagerConfig _managerConfig;
+
         [ObservableProperty]
         private string _quickPlayText = "Start Server and Connect";
+        
+        [ObservableProperty]
+        private ManagerConfig _managerConfig;
```
artehe, line 54: I just downloaded the branch and ran it. Everything seemed to work right even following the steps you did in the video it behaved properly.

Only thing I had to do was delete one of these as you have the same thing defined twice now

artehe (approved):

PR #76, opened 2024-03-06
Fix updater

Plootie:


SIT.Manager.Updater/Program.cs
```diff
@@ -90,8 +92,29 @@
...
+        {
+            entry.WriteToDirectory(releasePath.FullName, new ExtractionOptions()
+            {
+                ExtractFullPath = true,
+                Overwrite = true
+            });
+        }
+    }
+}
+
```
artehe, line 115: Now we found what's going wrong I think we leave using the SharpCompress package but this could be condensed down to this:
```csharp
using (ZipArchive archive = ZipArchive.Open(zipPath))
{
    archive.ExtractToDirectory(releasePath.FullName);
}
```

SIT.Manager.Avalonia/AssemblyInfo.cs
```diff
@@ -11,4 +11,8 @@
 [assembly: AssemblyCulture("")]
 [assembly: SupportedOSPlatform("windows")]
 [assembly: SupportedOSPlatform("linux")]
-[assembly: AssemblyVersion("0.0.0.1")]
\ No newline at end of file
+<<<<<<< Updated upstream
+[assembly: AssemblyVersion("0.0.0.0")]
+=======
+[assembly: AssemblyVersion("0.0.1.0")]
+>>>>>>> Stashed changes
```
artehe, line 18: ^^

Might want to fix this

artehe (changes requested):

SIT.Manager.Avalonia/AssemblyInfo.cs
Plootie: x.x

SIT.Manager.Updater/Program.cs
Plootie: true, true. I'll fix that ^^

artehe (approved):

PR #77, opened 2024-03-07
Backward version support

Plootie:


SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -29,6 +29,7 @@ public partial class PlayPageViewModel : ViewModelBase
         private const string EFT_EXE_FILENAME = "EscapeFromTarkov.exe";
         private readonly IManagerConfigService _configService;
         private readonly IServiceProvider _serviceProvider;
+        private static readonly Version standardUriFormatSupportedVersion = new Version("1.10.8827.30098");
```
artehe, line 32: It's just 1.10 and on that works doesn't it? so would it be better to just do ```new Version("1.10");```?

Plootie: I was looking at the commit for the slash fix that's [here](https://github.com/stayintarkov/StayInTarkov.Client/commit/c0879a12eb0b91ee7906b30667aaf9527c6008a9) which says it was pushed in release 1.10.8827.30098. And so if someone hasn't updated their 1.10 build yet it'll still error. Do you think we should just enforce the need to update SIT client?

artehe: Ah okay, I just assumed it was all 1.10 versions. It's fine as it is 😄

PR #102, opened 2024-03-17
Fix Opening SIT Config

Plootie:
Later versions of SIT use com.stayintarkov now. Not sure whether the use of a hashset is worth it looking back...


artehe:
Don't think it really matters tbh

PR #104, opened 2024-03-18
A nicer way of handling showing linux settings

artehe:
I just found out that you can do this which I think is nicer than how we currently show it.


Plootie (approved):
Had no clue that was a think. That makes some stuff super neat

artehe:
Is in here if you're interested: https://docs.avaloniaui.net/docs/guides/platforms/platform-specific-code/xaml

PR #106, opened 2024-03-19
Switch versioning to a different method

artehe:
Using this method should make updating for linux easier to setup as the verisons will be consistent on builds between Windows and Linux, unless they somehow are build just before and after midnight.

Example version is now 1.0.8837.39476

So we now have to manuall update the first 3 values but that shouldn't be too much effort to handle


artehe:
Hmm nope doesn't seem to work as good as I hoped 😢

PR #107, opened 2024-03-20
Activate views differently

artehe:
A problem I kept running into while doing the new install page was the way view activation was triggered. So I removed the dependency on ReactiveUI as that was pretty much all we used it for and added a way to do it using just CommunityMVVM.

Surprisingly the only thing this broke was the settings button translation in the main view, so converted the settings button to be able to be translated in XAML. Haven't seen any other problems with this but let me know 😄


Plootie (approved):
Haven't tested it but the code looks alll good

PR #108, opened 2024-03-21
Linux updating

### LINKED ISSUE #91: Integrate the Updater executable into main project
As we seem to be able to move the currently running executable in SIT.Manager.Avalonia.Desktop we could probably do away with the seperate Update exe project and integrate it directly into the main UI nicely.

This would save a load of space and make doing future releases easier to do

### LINKED ISSUE #92: Add a way to update for linux users
Taken from one of the TODOs left in the code but there isn't currently a way for Linux to update itself.

artehe:
Removes the updater project as it turns out we don't require a separate executable to manager updating for us, so a lot less to build and bundle together.

Also handle updating the Linux version of the app as well.

- Resolve #91
- Resolve #92


SIT.Manager.Avalonia/Services/AppUpdaterService.cs
```diff
@@ -0,0 +1,181 @@
...
+            try
+            {
+                Version currentVersion = Assembly.GetExecutingAssembly().GetName().Version ?? new Version("0");
+
+                string versionJsonString = await _httpClient.GetStringAsync(MANAGER_VERSION_URL);
+                GithubRelease? latestRelease = JsonSerializer.Deserialize<GithubRelease>(versionJsonString);
+                if (latestRelease != null)
+                {
+                    Version latestVersion = new(latestRelease.name);
+                    return latestVersion.CompareTo(currentVersion) > 0;
```
Plootie, line 92: I know this is back from the old manager. Can't we replace this with `latestVersion > currentVersion`?

Plootie (approved):

SIT.Manager.Avalonia/Services/AppUpdaterService.cs
artehe: Good shout lemme change that

artehe:
You might want to take another look at it, I kept finding bugs that needed fixing. It should now work however.

PR #113, opened 2024-03-27
Add new log generator button

Plootie:
Adds a new button on the tools page to generate a diagnostics zip containing the selected files

### DIFF
--- a/SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
+++ b/SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
@@ -34,7 +34,7 @@ public partial class ToolsPageViewModel : ObservableObject
     private readonly IFileService _fileService;
     private readonly ITarkovClientService _tarkovClientService;
     private readonly ILocalizationService _localizationService;
-    private readonly HttpClient _httpClient;
+    private readonly IDiagnosticService _diagnosticService;
     private static string EFTLogPath
     {
         get => Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "..", "LocalLow", "Battlestate Games", "EscapeFromTarkov", "Player.log");
@@ -58,7 +58,7 @@ public partial class ToolsPageViewModel : ObservableObject
                               IFileService fileService,
                               ILocalizationService localizationService,
                               ITarkovClientService tarkovClientService,
-                              HttpClient httpClient)
+                              IDiagnosticService diagnosticService)
     {
         _akiServerService = akiServerService;
         _barNotificationService = barNotificationService;
@@ -66,7 +66,7 @@ public partial class ToolsPageViewModel : ObservableObject
         _fileService = fileService;
         _tarkovClientService = tarkovClientService;
         _localizationService = localizationService;
-        _httpClient = httpClient;
+        _diagnosticService = diagnosticService;
 
         OpenEFTFolderCommand = new AsyncRelayCommand(OpenEFTFolder);
         OpenBepInExFolderCommand = new AsyncRelayCommand(OpenBepInExFolder);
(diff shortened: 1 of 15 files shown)


SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
```diff
@@ -217,4 +236,115 @@ private async Task ClearCache()
             }
         }
     }
+
+    private async Task GenerateDiagnosticReport()
+    {
+        var results = await new SelectLogsDialog().ShowAsync();
```
artehe, line 242: Would it make more sense if everything after here was in some kind of service, especially if it's likely that this could be added to/extended in the future?

SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
```diff
@@ -217,4 +236,115 @@ private async Task ClearCache()
...
+        }
+
+        //I hate doing it this way but at least icanhazip is owned by cloudlfare
+        HttpResponseMessage resp = await _httpClient.GetAsync("https://ipv4.icanhazip.com/");
+        string externalAddress = await resp.Content.ReadAsStringAsync();
+
+        List<Tuple<string, string>> CleanLogFiles = new(DiagnosticData.Count);
+        foreach (Tuple<string, string> diagnosticFile in DiagnosticData)
+        {
+            CleanLogFiles.Add(new(diagnosticFile.Item1, diagnosticFile.Item2.Replace(externalAddress.Trim(), "REACTED")));
```
artehe, line 308: I assume this meant to say "REDACTED"?

SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
```diff
@@ -217,4 +236,115 @@ private async Task ClearCache()
...
+                var entry = zipArchive.CreateEntry(entryData.Item1);
+                using (Stream entryStream = entry.Open())
+                using (StreamWriter sw = new(entryStream))
+                {
+                    await sw.WriteAsync(entryData.Item2);
+                    await sw.FlushAsync();
+                }
+            }
+        }
+        await fileStream.DisposeAsync();
```
artehe, line 348: Should fileStream not be as part of a using so that way it is always disposed of regardless what happens in here?

SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
```diff
@@ -217,4 +236,115 @@ private async Task ClearCache()
...
+    private async Task GenerateDiagnosticReport()
+    {
+        var results = await new SelectLogsDialog().ShowAsync();
+        List<Tuple<string, string>> DiagnosticData = new(4);
+
+        if (results.IncludeClientLog)
+        {
+            string eftLogPath = EFTLogPath;
+            DiagnosticData.Add(new(Path.GetFileName(eftLogPath), await File.ReadAllTextAsync(eftLogPath)));
+        }
```
artehe, line 249: Add a check to see if the file actually exists before you try and read it otherwise the app will close

SIT.Manager.Avalonia/Views/Dialogs/SelectLogsDialog.axaml
```diff
@@ -0,0 +1,21 @@
...
+				  Title="{DynamicResource ToolsSelectLogsTitle}"
+				  PrimaryButtonText="{DynamicResource ToolsSelectLogsProceedButtonText}"
+				  CloseButtonText="{DynamicResource ToolsSelectLogsCloseButtonText}"
+				  x:Class="SIT.Manager.Avalonia.Views.Dialogs.SelectLogsDialog"
+				  x:DataType="vm:SelectLogsDialogViewModel">
+	<StackPanel>
+		<CheckBox Name="ClientLogCheckbox" Content="{DynamicResource ToolsClientLogCheckboxContent}" IsChecked="{Binding SelectedOptions.IncludeClientLog, Mode=OneWay}" />
+		<CheckBox Name="ServerLogCheckbox" Content="{DynamicResource ToolsServerLogCheckboxContent}" IsChecked="{Binding SelectedOptions.IncludeServerLog, Mode=OneWay}" />
+		<CheckBox Name="DiagnosticLogCheckbox" Content="{DynamicResource ToolsDiagnosticsLogCheckboxContent}" IsChecked="{Binding SelectedOptions.IncludeDiagnosticLog, Mode=OneWay}" />
+		<CheckBox Name="HttpJsonCheckbox" Content="{DynamicResource ToolsHttpJsonCheckboxContent}" IsChecked="{Binding SelectedOptions.IncludeHttpJson, Mode=OneWay}" />
```
artehe, line 19: You'll probably want these to be Mode=TwoWay otherwise it'll always be true for every one of them

SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
artehe, line 308: One other point that's very minor is you could just do the trim once on external address rather than everytime in the foreach look

artehe (changes requested):

SIT.Manager.Avalonia/ViewModels/ToolsPageViewModel.cs
Plootie: nonono I TOTALLY meant to say "REACTED"

artehe (approved):

PR #115, opened 2024-03-28
Notify if there is an update available for SIT

### LINKED ISSUE #52: Check for available SIT updates
An idea I had would be to add a check somewhere to see if there is an update to the SIT client mod for example then provide a nice way to just update that provided that the version are compatible.

Could probably tie in nicely with #24

artehe:
Show a badge on the install page nav view if there is an update available for SIT which uses the same Tarkov version as currently installed.

image

Will then resolve #52

Not sure if you think it's worth including a button on the install page called 'Update to latest' which quickly updates them to the latest version users are able to get.

But have a little look at it and see what you think. I tested using EFT version 0.14.1.1.28965 and SIT version 1.10.8824.27027 installed


Plootie:
I think an update to latest button would be a great addition

artehe:
Thought you'd say that 😆

artehe:
I do believe it works properly now

Plootie (approved):

PR #116, opened 2024-03-28
Fixed Server Stacking and Deadlocking, translated Install Page, and more...

J0nathan550:
Fixed the stacking of the servers by hecking if the server is running in Play Page View Model.

Deadlocking was causing by the unhandled exception which user never get when he was closing the server (Probably because multiple servers runned at the same time in launcher).

Translated and updated missing keys in the in Install Page. Fixed not added "File-Service" translation.

Added ability to clear the console output if user wants it with additional check asking user if he wants to remove entire console output.

Fixed crocked visuals of the Install Page. Server Page.

Refactored few functions to make translation work, or fixed spelling mistake in one of the functions.

Removed useless http-client dependency. That was assigned, but never actually used.

### DIFF
--- a/SIT.Manager.Avalonia/Services/AkiServerService.cs
+++ b/SIT.Manager.Avalonia/Services/AkiServerService.cs
@@ -109,7 +109,7 @@ public class AkiServerService(IBarNotificationService barNotificationService,
 
     public override void Start(string? arguments = null)
     {
-        if (State == RunningState.Running)
+        if (State == RunningState.Running || State == RunningState.Starting)
         {
             return;
         }
--- a/SIT.Manager.Avalonia/Services/AkiServerService.cs
+++ b/SIT.Manager.Avalonia/Services/AkiServerService.cs
@@ -80,30 +80,24 @@ public class AkiServerService(IBarNotificationService barNotificationService,
 
     public bool IsUnhandledInstanceRunning()
     {
-        try
-        {
-            Process[] akiServerProcesses = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(SERVER_EXE));
+        Process[] akiServerProcesses = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(SERVER_EXE));
 
-            if (akiServerProcesses.Length > 0)
+        if (akiServerProcesses.Length > 0)
+        {
+            if (_process == null || _process.HasExited)
             {
-                if (_process == null || _process.HasExited) // error here (No process is associated with this object.)
+                return true;
+            }
+
+            foreach (Process akiServerProcess in akiServerProcesses)
+            {
+                if (_process.Id != akiServerProcess.Id)
                 {
                     return true;
                 }
-
-                foreach (Process akiServerProcess in akiServerProcesses)
-                {
-                    if (_process.Id != akiServerProcess.Id)
-                    {
-                        return true;
-                    }
-                }
             }
         }
-        catch
-        {
-            return false;
-        }
+
         return false;
     }
 
--- a/SIT.Manager.Avalonia/Services/AkiServerService.cs
+++ b/SIT.Manager.Avalonia/Services/AkiServerService.cs
@@ -80,24 +80,30 @@ public class AkiServerService(IBarNotificationService barNotificationService,
 
     public bool IsUnhandledInstanceRunning()
     {
-        Process[] akiServerProcesses = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(SERVER_EXE));
-
-        if (akiServerProcesses.Length > 0)
+        try
         {
-            if (_process == null || _process.HasExited)
-            {
-                return true;
-            }
+            Process[] akiServerProcesses = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(SERVER_EXE));
 
-            foreach (Process akiServerProcess in akiServerProcesses)
+            if (akiServerProcesses.Length > 0)
             {
-                if (_process.Id != akiServerProcess.Id)
+                if (_process == null || _process.HasExited) // error here (No process is associated with this object.)
                 {
                     return true;
                 }
+
+                foreach (Process akiServerProcess in akiServerProcesses)
+                {
+                    if (_process.Id != akiServerProcess.Id)
+                    {
+                        return true;
+                    }
+                }
             }
         }
-
+        catch
+        {
+            return false;
+        }
         return false;
     }
 
--- a/SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
+++ b/SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
@@ -19,7 +19,6 @@ public partial class InstallPageViewModel : ObservableRecipient,
     private List<InstallStep> _sitInstallSteps;
     private List<InstallStep> _serverInstallSteps;
     private readonly ILocalizationService _localizationService;
-    private readonly IManagerConfigService _managerConfigService;
 
     private InstallProcessState _installProcessState = new();
 
@@ -34,12 +33,12 @@ public partial class InstallPageViewModel : ObservableRecipient,
     [ObservableProperty]
     private ReadOnlyCollection<InstallStep> _installationSteps;
 
-    public InstallPageViewModel(ILocalizationService localizationService, IManagerConfigService managerConfigService)
+    public InstallPageViewModel(ILocalizationService localizationService)
     {
-        _managerConfigService = managerConfigService;
         _localizationService = localizationService;
-        _managerConfigService.ConfigChanged += (o, e) =>
+        _localizationService.LocalizationChanged += (o, e) =>
         {
+
             _sitInstallSteps =
             [
                 new(typeof(SelectView), _localizationService.TranslateSource("InstallPageViewModelSelectText")),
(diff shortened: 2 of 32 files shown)


Plootie (changes requested):
Please split this into different pull requests. There's too much going on in this

SIT.Manager.Avalonia/Services/AkiServerService.cs
```diff
@@ -80,30 +80,36 @@ public string[] GetCachedServerOutput()
...
+                        return true;
+                    }
+                }
             }
         }
-
+        catch
+        {
+            return false;
+        }
```
Plootie, line 106: These changes aren't needed anymore as the root issue is fixed

SIT.Manager.Avalonia/Services/AkiServerService.cs
```diff
@@ -80,30 +80,36 @@ public string[] GetCachedServerOutput()
...
         return false;
     }
 
     public override void Start(string? arguments = null)
     {
-        if (State == RunningState.Running)
+        if (State == RunningState.Running || State == RunningState.Starting)
```
Plootie, line 112: This is already pushed to master

SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
```diff
@@ -42,9 +34,45 @@ public partial class InstallPageViewModel : ObservableRecipient,
...
+            new(typeof(InstallView), _localizationService.TranslateSource("InstallPageViewModelInstallText")),
+            new(typeof(CompleteView), _localizationService.TranslateSource("InstallPageViewModelCompleteText"))
+        ];
+        _serverInstallSteps = 
+        [
+            new(typeof(SelectView), _localizationService.TranslateSource("InstallPageViewModelSelectText")),
+            new(typeof(ConfigureServerView), _localizationService.TranslateSource("InstallPageViewModelConfigureText")),
+            new(typeof(InstallView), _localizationService.TranslateSource("InstallPageViewModelInstallText")),
+            new(typeof(CompleteView), _localizationService.TranslateSource("InstallPageViewModelCompleteText"))
+        ];
```
Plootie, line 75: Not a fan of these updating on *any* config change

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -184,6 +177,9 @@ private async Task<string> LoginToServerAsync(Uri address)
                 CloseButtonText = _localizationService.TranslateSource("PlayPageViewModelButtonOk")
             }.ShowAsync();
         }
+        
+        QuickPlayText = _localizationService.TranslateSource("PlayPageViewModelQuickPlayText");
```
Plootie, line 181: Isn't this set elsewhere?

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -294,6 +290,10 @@ private async Task ConnectToServer(bool launchServer = false)
 
         if (launchServer)
         {
+            if (_akiServerService.State == RunningState.Starting)
+            {
+                return;
+            }
```
Plootie, line 296: This shouldn't return, otherwise the quick play button will do nothing if the server is started. Since the state check has been fixed in the aki service we can allow it to call start and wait for the server

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
```diff
@@ -342,14 +341,14 @@ private async Task ConnectToServer(bool launchServer = false)
...
-        string launchArguments = CreateLaunchArugments(new TarkovLaunchConfig { BackendUrl = backendUrl }, token);
+        string launchArguments = CreateLaunchArguments(new TarkovLaunchConfig { BackendUrl = backendUrl }, token);
         try
         {
             _tarkovClientService.Start(launchArguments);
         }
         catch (Exception ex)
         {
-            _logger.LogError("An exception occured while launching Tarkov: {exMessage}", ex.Message);
+            _logger.LogError($"An exception occured while launching Tarkov: {ex.Message}");
```
Plootie, line 351: This is the wrong format for the log. The compiler will actually get grumpy at you if you do this

SIT.Manager.Avalonia/Views/InstallPage.axaml
```diff
@@ -17,7 +17,7 @@
 		</Style>
 	</UserControl.Styles>
 					
-	<Grid ColumnDefinitions="Auto, *">
+	<Grid ColumnDefinitions="Auto, *" VerticalAlignment="Top" Margin="0,5,0,0">
```
Plootie, line 20: Not vibing with the vertical alignment being at the top instead of center

SIT.Manager.Avalonia/Views/ServerPage.axaml
```diff
@@ -53,28 +53,41 @@
...
+				<Button Name="ClearServerLogsButton"
+						HorizontalAlignment="Stretch"
+						Margin="0 0 10 0"
+						Grid.Column="2"
+						ToolTip.Tip="{DynamicResource ServerClearServerOutputTooltip}"
+						Command="{Binding ClearServerOutputCommand}">
+					<Grid ColumnDefinitions="auto,*">
+						<ui:SymbolIcon Symbol="Delete" Margin="0 0 10 0" Grid.Column="0"/>
+						<TextBlock Text="{DynamicResource ServerClearServerOutputText}" TextTrimming="CharacterEllipsis" Grid.Column="1"/>
+					</Grid>
```
Plootie, line 88: Why grid auto * over horizontal stack panel?

Plootie (changes requested):

SIT.Manager.Avalonia/Services/AkiServerService.cs
J0nathan550: I don't remember I ever done that I only remember I added try catch. Will fix it.

SIT.Manager.Avalonia/Services/AkiServerService.cs
J0nathan550: Merge issues :)

SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
J0nathan550: > Not a fan of these updating on _any_ config change

There is no other way to translate the steps.

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
J0nathan550: > Isn't this set elsewhere?

I removed that line when I was making the server fix, but with change of statement in the aki service I added it back.

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
J0nathan550: Forgot to remove, will remove.

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
J0nathan550: Sorry, will bring it back

SIT.Manager.Avalonia/Views/InstallPage.axaml
J0nathan550: When you extend the window to full screen it looks crappy at least for me. But will remove it.

SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
Plootie: Given how often this is an issue you could probably create a new event in the config service for when the localization changes

SIT.Manager.Avalonia/Views/ServerPage.axaml
J0nathan550: Because text that was showing to the user was breaking the visual, it's better to use that grid to first fix the visual and if space is not enough for visual it will show textblocks with trimming "..."

SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
J0nathan550: > Given how often this is an issue you could probably create a new event in the config service for when the localization changes

Good point

SIT.Manager.Avalonia/ViewModels/PlayPageViewModel.cs
J0nathan550: Was a wrong line placement. Put it where it belongs.

SIT.Manager.Avalonia/ViewModels/InstallPageViewModel.cs
```diff
@@ -42,9 +33,45 @@ public partial class InstallPageViewModel : ObservableRecipient,
...
```
artehe, line 74: Just a note so you guys know I think this is just gonna cause so much more hassle in the future than it's worth (and you can kind of see already if you open the app, try a server install, and  look closely at the stepper and you'll probably see part of what I mean) with dynamically changing the Stepper values like this, so at some point I'll change probably them to just be in XAML.

PR #121, opened 2024-03-29
Fix deadlock

Plootie:


J0nathan550:
Is it fixed or not yet?

Plootie:
This does stop the deadlock but i literally have no idea why the `Close` call was deadlocking. Probably something to do with redirecting the output stream

SIT.Manager.Avalonia/ManagedProcess/ManagedProcess.cs
```diff
@@ -48,10 +48,12 @@ public virtual void Stop()
...
-
-        _process.WaitForExit();
-        _process.Close();
+        if (clsMsgSent)
+        {
+            if (!_process.WaitForExit(TimeSpan.FromSeconds(5)))
+            {
+                _process.Kill();
+            }
+        }
```
artehe, line 57: 'clsMsgSent' should be always false if close after launch isn't on though is it not meaning this will never actually run?

Would something like this work better?

```csharp
        bool closed = false;

        // Stop the server process
        bool clsMsgSent = _process.CloseMainWindow();
        if (clsMsgSent)
        {
            closed = _process.WaitForExit(TimeSpan.FromSeconds(5));
        }

        if (!closed)
        {
            _process.Kill();
            _process.WaitForExit(TimeSpan.FromSeconds(5));
        }
 
        _process.Close();

```

Plootie: You are absolutely correct. Writing code while sleep deprived gives interesting results

PR #125, opened 2024-03-30
Rename project to SIT.Manager

Plootie:
Wanna make sure this doesn't absolutely break everything. It's building and running for me but want the confirmation


SIT.Manager/Services/AppUpdaterService.cs
```diff
@@ -31,11 +31,11 @@ private static string ProcessName
         {
             if (OperatingSystem.IsWindows())
             {
-                return "SIT.Manager.Avalonia.Desktop.exe";
+                return "SIT.Manager.Desktop.exe";
             }
             else if (OperatingSystem.IsLinux())
             {
-                return "SIT.Manager.Avalonia.Desktop";
+                return "SIT.Manager.Desktop";
```
artehe, line 38: Are these executable names going to be right?

SIT.Manager/Services/AppUpdaterService.cs
```diff
@@ -14,11 +14,11 @@
 using System.Text.Json;
 using System.Threading.Tasks;
 
-namespace SIT.Manager.Avalonia.Services;
+namespace SIT.Manager.Services;
 
 public class AppUpdaterService(IFileService fileService, ILogger<AppUpdaterService> logger, HttpClient httpClient, IManagerConfigService managerConfigService) : IAppUpdaterService
 {
-    private const string MANAGER_VERSION_URL = @"https://api.github.com/repos/stayintarkov/SIT.Manager.Avalonia/releases/latest";
+    private const string MANAGER_VERSION_URL = @"https://api.github.com/repos/stayintarkov/SIT.Manager/releases/latest";
```
artehe, line 21: Is this URL still going to be correct?

SIT.Manager/Services/AppUpdaterService.cs
```diff
@@ -47,11 +47,11 @@ private static string ReleaseUrl
         {
             if (OperatingSystem.IsWindows())
             {
-                return @"https://github.com/stayintarkov/SIT.Manager.Avalonia/releases/latest/download/win-x64.zip";
+                return @"https://github.com/stayintarkov/SIT.Manager/releases/latest/download/win-x64.zip";
             }
             else if (OperatingSystem.IsLinux())
             {
-                return @"https://github.com/stayintarkov/SIT.Manager.Avalonia/releases/latest/download/linux-x64.tar";
+                return @"https://github.com/stayintarkov/SIT.Manager/releases/latest/download/linux-x64.tar";
```
artehe, line 54: As previous

artehe (commented):
Mainly just those queries around updating and urls/executable names they'll use but otherwise don't think they'll be a problem

SIT.Manager/Services/AppUpdaterService.cs
Plootie: Thanks for that catch. I ran the find and replace before I decided to rename the executable

SIT.Manager/Services/AppUpdaterService.cs
Plootie: Nope... That's what I can for running a find and replace without checking,,,

PR #132, opened 2024-03-31
Add basic crash warning

Plootie:
I'm debating whether I want to implement a localization system into this or avoid it to minimize the risk of a bug in the localization service interfering with the crash warning (throw exception left in for testing)

### DIFF
--- a/SIT.Manager/SIT.Manager.csproj
+++ b/SIT.Manager/SIT.Manager.csproj
@@ -47,9 +47,6 @@
             <AutoGen>True</AutoGen>
             <DependentUpon>Resources.resx</DependentUpon>
         </Compile>
-        <Compile Update="Views\CrashWindow.axaml.cs">
-          <DependentUpon>CrashWindow.axaml</DependentUpon>
-        </Compile>
     </ItemGroup>
 
     <ItemGroup>
--- a/SIT.Manager/SIT.Manager.csproj
+++ b/SIT.Manager/SIT.Manager.csproj
@@ -47,6 +47,9 @@
             <AutoGen>True</AutoGen>
             <DependentUpon>Resources.resx</DependentUpon>
         </Compile>
+        <Compile Update="Views\CrashWindow.axaml.cs">
+          <DependentUpon>CrashWindow.axaml</DependentUpon>
+        </Compile>
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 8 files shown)


SIT.Manager/Views/MainView.axaml.cs
```diff
@@ -16,6 +16,8 @@ public MainView()
     {
         InitializeComponent();
 
+        throw new Exception();
+
```
artehe, line 20: Well this certainly makes it crash 😉

SIT.Manager/SIT.Manager.csproj
```diff
@@ -47,6 +47,9 @@
             <AutoGen>True</AutoGen>
             <DependentUpon>Resources.resx</DependentUpon>
         </Compile>
+        <Compile Update="Views\CrashWindow.axaml.cs">
+          <DependentUpon>CrashWindow.axaml</DependentUpon>
+        </Compile>
```
artehe, line 52: This seems to make Visual Studio look off for me and don't think it needs it

SIT.Manager/Views/CrashWindow.axaml
```diff
@@ -0,0 +1,36 @@
...
+        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+		xmlns:vm="clr-namespace:SIT.Manager.ViewModels"
+        mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
+        x:Class="SIT.Manager.Views.CrashWindow"
+		x:DataType="vm:CrashWindowViewModel"
+        MaxHeight="240"
+		MaxWidth="450"
+		Title="Uh oh!"
+		WindowStartupLocation="CenterScreen">
+	<Grid RowDefinitions="auto, *, auto">
```
artehe, line 13: Probably want to add this into the Window definition: ```Icon="/Assets/Stay-In-Tarkov-512.png"```

artehe (commented):
Can probably just roll with this without any localisation for now. 

Can always add it later if we end up with a tonne of crashes, but like you said probably better to not have much more than we need for this as it should only appear if we've screwed up and not handled something properly.

SIT.Manager/SIT.Manager.csproj
Plootie: This was generated by visual studio and I have no idea what it does

PR #133, opened 2024-03-31
Downpatcher request bandaid

Plootie:


SIT.Manager/Services/Install/InstallerService.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Microsoft.Extensions.Logging;
+﻿using Avalonia.Labs.Controls;
```
artehe, line 1: Won't make much difference to us currently but the ContentDialog you create in here come from this package rather than "using FluentAvalonia.UI.Controls;" like it does everywhere else

Plootie: Blame visual stupid for that one c:

Issue #142, opened 2024-03-31
No localization in RELEASE BUILD


J0nathan550:
://///////////

![image](https://github.com/stayintarkov/SIT.Manager.Avalonia/assets/88881607/c97c3427-2b95-4c77-872f-592d845f0a03)

PR #144, opened 2024-03-31
Just fixed localization not loading in the settings.

J0nathan550:
One line of code, solves millions of issues.
Resolving: #142


Plootie (changes requested):
Move the name into a string constant please

Plootie (approved):

Issue #110, opened 2024-03-24
Remove the need for the extra files flag in the Mods Page


artehe:
The "Required Extra Files" flag for modding in the manager, it seems like a lot of people don't really understand the process.

So I think the best approach would be to include the dependecy dll into the https://github.com/stayintarkov/SIT-Mod-Ports for those that require any porting but others that don't need can just be linked to.

Can replace the "Required Extra Files" flag with a section of dependencies that either links to the reuqired download files or the other required mod in the master list. Then the manager can just install all of these dependencies as well.

Plootie:
@artehe Given the impending changes to mods do we want to close this issue?

artehe:
Ye that probably makes sense

Issue #103, opened 2024-03-18
Add recommended mod installs to installer


artehe:
When the new installer is made, we should implement recommended additional mods somehow (Configuration Manager pretty much) as a default as everyone ends up using them anyway.

PR #153, opened 2024-04-02
WIP: Added functionality to install additional mod files (required files)

Beat2er:
- Downloads zip, extracts it and detects client and server folder and copies all files (except ones in `PluginFiles` by SIT)
- Needs a new entry in MasterList.json for each mod like `"OriginalDownloadUrl": "https://github.com/TommySoucy/MoreCheckmarks/releases/download/v1.5.11/MoreCheckmarks-1.5.11-for-3.8.0-29197.zip",` or to `hub.sp-tarkov.com/file`
  - Button greyed out if `OriginalDownloadUrl` is missing
- <s>[ ] translations</s> (help wanted)
- [x] created translation keys for all languages but only filled with english content
- Closes: #110

\
I'm open to feedback, whether the project would be interested in this PR at all and how the translations are done.
Currently I am still fixing some edge-cases:
- <s>[ ] client only</s> (No need as you can't remove aki SPT_AKI path once set)
- <s>[ ] download from `hub.sp-tarkov.com/file` once it's fully up again</s> (Post request required to get the link)
- [x] <s>explore possibility for</s> dynamic download link based on version (support github and gitea hosted on dev.sp-tarkov.com to dynamically get appropriate release asset)
- [x] check if 7z etc also working (made a small addition for extracting 7z, but is rarely used)
- [x] basic dependencies (only of listed mods)
- [x] cache/save resolution of github and gitea releases, as api rate limits apply
- [ ] more testing


artehe:
Hey, thanks for doing this saves me a job! 😄 

I'll have a look through it all properly when you think it's ready.

The issue to link it to is #110 as we basically want to make the required install flag redundant so that installing mods works better rather than add things like the configuration manager as part of the SIT install a toggalable option.

I was thinking it might be an idea to have it look into dealing with dependencies which could be in the master mod list as well due to them needing some kind of porting as well.

You can just leave this PR as handling client only mods dependencies for now, we can do another PR later which works out server mod installing as the manager doesn't do that at all yet.

To answer the question about translations if you're adding new ones then I would put all the same keys and their respective english into all the other localisation files as a minimum.

Beat2er:
Hi, yea. 
I forgot about the dependencies, will add an early version.
My general plan is to have a working mvp, that helps less tech-savvy people without overcomplicating it. Enought to be a good first iteration.
I'll give you a head's up when I am pleased.
Thanks

artehe:
Sounds good to me, mod installation was the next big target as it was a cause of a lot of questions.

Some other things I thought of which you might want to bear in mind. 
- At some point I'll completly change the way the Mods page looks (unless someone beats me to it), so that it looks/feels a lot better and would make more sense showing what's already installed, it's version, what mods are available and what if any potential updates there are etc etc.
- Also on top of that will be server mods eventually and fitting that in UI wise so it's the same but also seperate enough that it's clear. Although this will probably come after everything else as getting the client side stuff working is the more important/harder one due to the need for specific porting with certain mods to be compatible.

So tackling the required extra files part is a very useful step! ❤️

artehe:
Just updated your issue text quickly so it linked to the right issue. 

I finally got around to doing something for #103, as they're both going to have a lot of changes to the ModService I'll won't merge mine till after this is done so you won't have to deal with the horrific merge conflict that will happen.

Beat2er:
will probably finish up tomorrow

cookingboy3:
Out of curiosity, is there any reason why each mod doesn't define a mod.json file with it's version and the files that need to be copied to the server and client? Using the master list rather than enumerating each mod individually creates a big merge conflict point as well as closing the door for launcher support of third-party ports of mods.

Plootie:
The way we handle mods and subsequently the mods page is changing for 1.11 and as such this PR will no longer be needed. Thank you for starting it though!

PR #158, opened 2024-04-03
Add recommended mod install

### LINKED ISSUE #103: Add recommended mod installs to installer

artehe:
Resolves #103 

Will add the option to install certain mods when installing SIT itself, to start with this will just be the Configuration Manager as most people will want this anyway.


artehe:
I think this is good to go now?

SIT.Manager/Services/ModService.cs
```diff
@@ -27,23 +28,75 @@ public class ModService(IBarNotificationService barNotificationService,
...
         }
-        Directory.CreateDirectory(Path.Combine(modsDirectory, "Extracted"));
+    }
+
+    public async Task DownloadModsCollection()
+    {
+        if (!Directory.Exists(_localModCache))
+        {
+            Directory.CreateDirectory(_localModCache);
+        }
```
Plootie, line 91: We can shorten this to just `Directory.CreateDirectory(_localModCache);` as the create directory method doesn't throw on already existing

SIT.Manager/ViewModels/Installation/ConfigureSitViewModel.cs
```diff
@@ -174,6 +182,39 @@ private void ValidateConfiguration()
...
+
+        // Make sure that all the recommended mods are selected to start with - this isn't the best way of doing it but it works ¯\_(ツ)_/¯
+        List<ModInfo> selectedMods = [];
+        foreach (ModInfo mod in Mods)
+        {
+            if (_modService.RecommendedModInstalls.Contains(mod.Name))
+            {
+                selectedMods.Add(mod);
+            }
+        }
```
Plootie, line 214: I dunno if im sleep deprived but i'm confused by this

SIT.Manager/Services/ModService.cs
artehe: Good point

SIT.Manager/ViewModels/Installation/ConfigureSitViewModel.cs
artehe: So am I. I think I was just having a special moment 😓 

Have used something that makes more sense now.

Plootie (approved):

PR #161, opened 2024-04-04
Windows edition SIT.Manager compile

DRDRDRRDDRDR:
Original .yml code by the true hero [NuckedBart](https://github.com/NukedBart/SIT.Manager.Avalonia/tree/master) 
Changed solution name and -Path to fits the name change


LIUZHE8:
the update history is so EMBARRASSING

mihaicm93:
> the commit history is so EMBARRASSING

we do not judge, we are all learning here

Plootie:
Believe me, it's better than mine when I attempted this ;P
I'm really appreciative of you taking time to do this for us but I'm not sure if we'll be able to use this as is >.<

The manager's auto-update system uses its assembly version to compare against the latest github release version but the wildcard versioning uses seconds since midnight divided by 2 for the 4th position. That means if you aren't using `-property:Version=x.x.x.x` in your build command and they aren't both compiled at the EXACT same time then one of the builds will think it's always out of date

DRDRDRRDDRDR:
> Believe me, it's better than mine when I attempted this ;P I'm really appreciative of you taking time to do this for us but I'm not sure if we'll be able to use this as is >.<
> 
> The manager's auto-update system uses its assembly version to compare against the latest github release version but the wildcard versioning uses seconds since midnight divided by 2 for the 4th position. That means if you aren't using `-property:Version=x.x.x.x` in your build command and they aren't both compiled at the EXACT same time then one of the builds will think it's always out of date

Thank you so much,i'll fix it asap

DRDRDRRDDRDR:
return master to the original style,sry for the spam
try new stuff in a branch later
i'm a pure newbie

artehe:
> return master to the original style,sry for the spam try new stuff in a branch later i'm a pure newbie

Don't worry about it as mihai said we don't judge or mind, everyone is learning. Just take your time and have fun 😄

mihaicm93:
I think I had a PR with 90 commits when I was testing the workflow for package and build the server + server mod, there is a way to test them locally but requires a bit of set-up, it's just easier to do it on github directly

DRDRDRRDDRDR:
alright i realized what's going on with the `-property:Version=x.x.x.x` command 
I thought it can change the version number automatically
but when it's been add in the compile command,MSbuild will choose the version `x.x.x.x`
so I'm confused about the way to reach the goal

artehe:
Had a look today at versioning and dear me I knew it was a massive pain but didn't think it would be quite as awkward as it actually is.

Made a PR which should make your life a lot easier #163 

So when that gets merged all you should need to do is call something like this for a build:
```dotnet build -c Debug -p:BuildNumber=333 -p:RevisionNumber=222```

Or if you want to make a release:
```dotnet publish -c Release -r win-x64 -p:BuildNumber=333 -p:RevisionNumber=222```

Then you can just generate something for the BuildNumber and RevisionNumber as Ploot mentioned earlier the manager currently sets what would be the BuildNumber to the number of days since the year 2000 and the RevisionNumber to be the number of seconds since midnight (divided by 2). We're not strictly tied to these and I don't think we particularly mind what they are, the main thing would be that they are consistent if we did a release so that Windows and Linux would get the same version numbers.

Hope this helps!

DRDRDRRDDRDR:
ALL DONE! 
Thanks to all of your guys for the help!
图片

artehe:
Thanks for setting this up 😄

PR #162, opened 2024-04-04
Concurrency and Game Copy Validation

artehe:
Not sure how well this is actually working so am open to opinions on this.

Should make the download and extraction for the patcher run at the same time as copying the EFT directory if either are required. Will also ensure validate that any eft files already copied are correct so that it won't copy if they aren't required.

From testing however this seems to run a lot slower than it did before so I might be being quite stupid in what/how I've done things.

### DIFF
--- a/SIT.Manager/Interfaces/IFileService.cs
+++ b/SIT.Manager/Interfaces/IFileService.cs
@@ -5,7 +5,7 @@ namespace SIT.Manager.Services;
 
 public interface IFileService
 {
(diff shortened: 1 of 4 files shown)


Plootie:
Might be even better solutions than this but this is my immediate gut response. In order to prevent this slowing things down we could store a csv/json file of the file path, its hash and its last write time value. When we go to do a subsequent copy we can compare write times and (in theory at least) if the file hasn't been modified we can skip checking the hash. We could the provide a separate "validate" feature that does check the hashes. Best of both worlds or stupid idea? You decide

PR #164, opened 2024-04-05
Add support for BepInEx patchers

caellach:
Adds support for patchers
Fixes 2 bad config paths usage

Related changes: https://github.com/stayintarkov/SIT-Mod-Ports/pull/38


artehe (approved):
Good catch on those two

PR #167, opened 2024-04-07
Hide mods page in devloper mode

artehe:
When you enable developer mode make the mods tab un-navigatable


mihaicm93:
I have no idea if it would make sense to add it to this PR or make another but would be good to have a check for the plugins folder if there is any other .dll's except stayintarkov.dll and configmanager.dll to get an error msg or just remove the extra mods

artehe:
> I have no idea if it would make sense to add it to this PR or make another but would be good to have a check for the plugins folder if there is any other .dll's except stayintarkov.dll and configmanager.dll to get an error msg or just remove the extra mods

I was planning on doing this at some point, the part I can't decide on is where to put it at the moment. 

I'm leaning towards running the check itself on when you enable developer mode have it yell at you if there's any mods it detects as installed and tells you to get rid of them first.

Plootie:
> I'm leaning towards running the check itself on when you enable developer mode have it yell at you if there's any mods it detects as installed and tells you to get rid of them first.

IMO this is the best solution

artehe:
> IMO this is the best solution

Cool, I'll do that then.

It'll be reliant on #158 as I can use the updated mod service to check if they have mods we don't mind them using while testing SIT builds such as configuration manager. 

We can also add the same check onto the connect button or whatever it'll be in the play page as well, but that can wait till @Plootie has finished with the UI changes there. Just to really make sure users can accidentally do something we don't want them to.

artehe:
@mihaicm93 @Plootie 

I think this is working happily now feel free to test it and see what you think should have the additional check to stop users enabling developer mode with any mods apart from Configuration manager installed.

Plootie:
👍I'll give this a test soon :3

mihaicm93:
Works great, I will have to get a Pre-release to try it fully with that but otherwise all good on my end
👍

mihaicm93 (approved):

SIT.Manager/ViewModels/MainViewModel.cs
```diff
@@ -70,7 +73,14 @@ public partial class MainViewModel : ObservableRecipient, IRecipient<Installatio
...
-        _managerConfigService.ConfigChanged += async (o, c) => await CheckForUpdate();
+        _managerConfigService.ConfigChanged += ManagerConfigService_ConfigChanged;
+    }
+
+    private async void ManagerConfigService_ConfigChanged(object? sender, ManagerConfig e)
+    {
+        IsDevloperModeEnabled = e.EnableDeveloperMode;
+
+        await CheckForUpdate();
     }
```
Plootie, line 84: Maybe silly question. Why do we check for an update when toggling dev mode and if this is needed will caching get in the way of this working properly?

Plootie (commented):
Just need to perform some tests with this but the code looks good. Just this one question

Plootie:
Found a pretty hilarious bug that I don't even know how to approach honestly. When enabling the developer mode we hide the mods page, however if you navigate to the mod page first, then enable the dev mode and use the "browser back" buttons on your mouse you can navigate back to the mods page while in dev mode. Idk how many people would ever actually do this but might be worth trying to stop

mihaicm93:
That is crazy, so I guess a check before game on the plugins folder?

artehe:
> Found a pretty hilarious bug that I don't even know how to approach honestly. When enabling the developer mode we hide the mods page, however if you navigate to the mod page first, then enable the dev mode and use the "browser back" buttons on your mouse you can navigate back to the mods page while in dev mode. Idk how many people would ever actually do this but might be worth trying to stop

Well that's unexpected. I think I found the cause those

SIT.Manager/ViewModels/MainViewModel.cs
artehe: That's been there the whole time, I just stuck ```IsDevloperModeEnabled = e.EnableDeveloperMode;``` in there as that was the most appropriate place for it.

probably should look into that separately as it does currently run the app update check every time that there is a ManagerConfig change. It does also check for a SIT update but that only happens once every hour

PR #168, opened 2024-04-09
Create Release action

artehe:
Create an action which will generate a release for us so no more building anything locally and is now just the click of a button😄

Also:
- Switch the current action over to use linux so that it works quicker 
- Get the current action building for both linux and windows.
- Have the current action only build debug versions as they are all we really need from it

### DIFF
--- a/.github/workflows/Create-Release.yml
+++ b/.github/workflows/Create-Release.yml
@@ -38,4 +38,4 @@ jobs:
         Write-Output "VERSION=$BuildTime" >> $env:GITHUB_OUTPUT 
         
     - name: Build Published Version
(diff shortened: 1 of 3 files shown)


Plootie (approved):
I have literally no clue how you review changes to github workflows but I'll take your word that this works :3

DRDRDRRDDRDR:
that's my fault XD

artehe:
> that's my fault XD

You gave me a really good basis to make the Release action so we can push releases to here https://github.com/stayintarkov/SIT.Manager.Avalonia/releases whenever we want without having to build anything ourselves locally etc.

Then the other changes I made to yours were just tweaks that I found when making the release action like realizing it'd be quicker running on Linux (saves minutes as we only have so many).

PR #169, opened 2024-04-09
Fix Update Url

artehe:
Use a link that will actually work for linux


Plootie (approved):
Oops, i forgot we started pushing tar.gz over straight tar

PR #172, opened 2024-04-11
Allow users to use default mirror on download

jubiman:
Added a simple ValidateConfiguration call at the end of FetchVersionAndMirrorMatrix because it was bugging me :)
This now allows users to simply click next on the install instead of having to change mirror (even if the mirror they want to use is selected).


artehe (approved):
No idea how I never noticed this before. 😓 

Thanks for doing this!

PR #180, opened 2024-04-12
Properly fix configmanager installation

artehe:
Create the directory that it should rather than creating a directory of the filename

Currently creates the directory below rather than just ensuring that the one it's trying to copy to exists:
image


Plootie:
oops..

PR #181, opened 2024-04-12
Caching in some places

artehe:
Wrapped up a couple of things into this

1. Cache the last check time for manager update checking and only do this once an hour. Does potentially mean a user could start the manager get told there is an update, restart the manager and then not get told about the update at the moment but will dramatically reduce the amount of usage of the api. However might mean we want a button somewhere that forces a check or something?
2. Fix the SIT update check so that it actually will check for an update now.
3. Cache the json serializer options for the manager config as we use them a lot so no need to recreate it constantly.
4. Should hopefully fix the weird issue I saw earlier when trying to install mods.

### DIFF
--- a/SIT.Manager/Services/AppUpdaterService.cs
+++ b/SIT.Manager/Services/AppUpdaterService.cs
@@ -97,26 +97,31 @@ public class AppUpdaterService(IFileService fileService, ILogger<AppUpdaterServi
 
     public async Task<bool> CheckForUpdate()
     {
-        if (_managerConfigService.Config.LookForUpdates)
+        Version currentVersion = Assembly.GetEntryAssembly()?.GetName().Version ?? new Version("0");
+        Version latestVersion = new();
+
+        if (_managerConfigService.Config.LookForUpdates && DateTime.Now.AddHours(-1) > _managerConfigService.Config.LastManagerUpdateCheckTime)
         {
             try
             {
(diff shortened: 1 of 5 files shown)


SIT.Manager/Services/AppUpdaterService.cs
```diff
@@ -97,26 +97,31 @@ private async Task ExtractUpdatedManager(string zipPath, string destination)
 
     public async Task<bool> CheckForUpdate()
     {
-        if (_managerConfigService.Config.LookForUpdates)
+        Version currentVersion = Assembly.GetEntryAssembly()?.GetName().Version ?? new Version("0");
+        Version latestVersion = new();
+
+        if (_managerConfigService.Config.LookForUpdates && DateTime.Now.AddHours(-1) < _managerConfigService.Config.LastManagerUpdateCheckTime)
```
Plootie, line 103: This code kinda threw me through a bit of a mental loop for a moment. Could we instead do something like `DateTime.Now - _managerConfigService.Config.LastManagerUpdateCheckTime > TimeSpan.FromHours(1)`?

Plootie (approved):
All good other than this one bit that broke my small brain a little

SIT.Manager/Services/AppUpdaterService.cs
artehe: Ye it makes me have to think every time but we're looking for when the last update time is over an hour ago and I'm not sure of a more obvious way of doing it

PR #182, opened 2024-04-12
Linux settings and wine DLL overrides (so tarkov doesn't call the police on you)

### LINKED ISSUE #178: Issues running SIT on linux
Whenever I get in a raid on linux, I get openGL errors: 
```
01d0:err:d3d:wined3d_debug_callback 0000000069631720: "GL_INVALID_OPERATION in glUseProgram(transform feedback active)".
01d0:err:d3d:wined3d_debug_callback 0000000069631720: "GL_INVALID_OPERATION in glDispatchCompute(no active compute shader)".
```

To be sure I downloaded the SPT Aki installer for linux using Lutris on the same prefix, and that works fine:

I'm not sure if this is an issue with the way tarkov is launched from the manager, or if it is an issue with the SIT client mod.
I have tried a lot of environment variables and I also copied some arguments from the official SPT Aki launcher repo, but nothing has worked so far.

jubiman:
Adds an entire new page for linux settings and launch options for wine. Eventually the Linux Settings portion in the settings will be moved here as well

Closes #178


Plootie:
A lot of the current linux code goes over my head so I'm gonna ping @artehe for a review whenever he's free. I'll read over what you've got but I can only really comment on the code itself and no the code design :3

SIT.Manager/Services/Install/InstallerService.cs
```diff
@@ -759,4 +783,16 @@ public void CopyEftSettings(string targetInstallDir)
...
+    {
+        string bsgLauncherPath = Path.Combine(_configService.Config.InstallPath, "BSGLauncher.exe");
+        if (File.Exists(bsgLauncherPath))
+        {
+            return;
+        }
+
+        string bsgLauncherUrl = "https://prod.escapefromtarkov.com/launcher/download";
+        await _fileService.DownloadFile("BsgLauncher.exe", _configService.Config.InstallPath, bsgLauncherUrl, null);
+    }
```
Plootie, line 797: Can find any references to this method but the general sentiment is that we should avoid touching BSGs stuff

Plootie (commented):
Haven't had a chance to go over all the files yet but I'll post my current comments until i find time for more

SIT.Manager/Controls/EmbeddedProcessWindow.cs
```diff
@@ -102,6 +102,11 @@ public async Task StartProcess()
                 // Set ProcessWindowHandle to the MainWindowHandle of the process
                 ProcessWindowHandle = _p.MainWindowHandle;
             }
+            else // linux
+            {
+                // Wait until the process has exited
+                await _p.WaitForExitAsync();
+            }
```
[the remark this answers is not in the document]
jubiman: This is some older stuff I added when trying to improve the installer, which I will finish at a later date.
I get an error because  [PatchViewModel.cs#L99](https://github.com/stayintarkov/SIT.Manager.Avalonia/blob/4238bf66ca0f75f86b4ed2cb7f7afa121996167d/SIT.Manager/ViewModels/Installation/PatchViewModel.cs#L99) only starts the process and doesn't wait for it to finish, but does then try to read the exit code. This was a fix for it but it should probably be in PatchViewModel instead

SIT.Manager/Services/Install/InstallerService.cs
jubiman: Just some lingering stuff to make the install process work just like lutris' (install the official launcher, log in, download EFT, close the launcher and then proceed with the normal SIT installation process). If you'd rather have people install EFT through lutris first and then having to install SIT, that would be possible, but in terms of UX that would kinda suck

Plootie: Yeah I understand it would suck to have them do that but as a project we like to avoid stepping on BSGs toes at all to avoid creating problems

SIT.Manager/Services/ManagerConfigService.cs
```diff
@@ -12,15 +12,20 @@ internal sealed class ManagerConfigService : IManagerConfigService
...
 
-    private ManagerConfig _config = new();
-    public ManagerConfig Config
-    {
-        get => _config;
-        private set { _config = value; }
-    }
+    public ManagerConfig Config { get; private set; } = OperatingSystem.IsLinux()
+        ? new LinuxConfig()
+        : new ManagerConfig();
```
Plootie, line 17: I think I'd like @artehe 's opinion on this. Idk if I groove with doing things this way as it would mean any linux code needing to access linux options will need to cast the manager config to a linux config beforehand

SIT.Manager/Services/TarkovClientService.cs
```diff
@@ -77,29 +79,77 @@ public override void Start(string? arguments)
...
+            // Check if either mangohud or gamemode is enabled.
+            if (config.IsGameModeEnabled)
+            {
+                _process.StartInfo.FileName = "gamemoderun";
+                _process.StartInfo.Arguments = string.Empty;
+                if (config.IsMangoHudEnabled)
+                {
+                    _process.StartInfo.Arguments += "\"mangohud\"";
+                }
+                _process.StartInfo.Arguments += $" \"{config.WineRunner}\"";
```
Plootie, line 93: Similar to my previous comment about process arguments

SIT.Manager/Services/TarkovClientService.cs
```diff
@@ -77,29 +79,77 @@ public override void Start(string? arguments)
...
+            _process.StartInfo.EnvironmentVariables.Add("__GL_SHADER_DISK_CACHE_PATH", winePrefix);
+            _process.StartInfo.EnvironmentVariables.Add("DXVK_STATE_CACHE_PATH", winePrefix);
+            // TODO: configure these with the DLLManager and add the ability to add custom DLL overrides.
+            // TODO: Something in these DLL overrieds fixes the graphical issues with the game. Time to figure out which ones
+            //_process.StartInfo.EnvironmentVariables.Add("WINEDLLOVERRIDES", "\"d3d10core,d3d11,d3d12,d3d12core,d3d9,d3dcompiler_33,d3dcompiler_34,d3dcompiler_35,d3dcompiler_36,d3dcompiler_37,d3dcompiler_38,d3dcompiler_39,d3dcompiler_40,d3dcompiler_41,d3dcompiler_42,d3dcompiler_43,d3dcompiler_46,d3dcompiler_47,d3dx10,d3dx10_33,d3dx10_34,d3dx10_35,d3dx10_36,d3dx10_37,d3dx10_38,d3dx10_39,d3dx10_40,d3dx10_41,d3dx10_42,d3dx10_43,d3dx11_42,d3dx11_43,d3dx9_24,d3dx9_25,d3dx9_26,d3dx9_27,d3dx9_28,d3dx9_29,d3dx9_30,d3dx9_31,d3dx9_32,d3dx9_33,d3dx9_34,d3dx9_35,d3dx9_36,d3dx9_37,d3dx9_38,d3dx9_39,d3dx9_40,d3dx9_41,d3dx9_42,d3dx9_43,dxgi,nvapi,nvapi64=n;winemenubuilder=");
+            string str = DllManager.GetDllOverride(config);
+            Console.WriteLine(str);
```
Plootie, line 136: Little debugging left in ;P

SIT.Manager/ViewModels/Installation/InstallationViewModelBase.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Messaging;
 using SIT.Manager.Models.Installation;
+using System;
```
Plootie, line 4: ? ;P

SIT.Manager/ViewModels/LinuxSettingsPageViewModel.cs
```diff
@@ -0,0 +1,114 @@
...
+    private readonly IManagerConfigService _configsService;
+    private readonly IPickerDialogService _pickerDialogService;
+    private readonly IBarNotificationService _barNotificationService;
+    private readonly ILocalizationService _localizationService;
+    
+    [ObservableProperty]
+    private LinuxConfig _config;
+    // DXVK Versions TODO
+    [ObservableProperty] 
+    private List<string> _dxvkVersions;
```
Plootie, line 26: `private List<string> _dxvkVersions = []`?

SIT.Manager/ViewModels/LinuxSettingsPageViewModel.cs
```diff
@@ -0,0 +1,114 @@
...
+
+    private void DeleteEnv()
+    {
+        // TODO: get the selected item and remove it from the dictionary
+    }
+
+    private void AddEnv()
+    {
+        Config.WineEnv.Add("", "");
+    }
```
Plootie, line 65: ;P

Plootie (commented):
Artehe probably has better insight into a lot of this than I will but mostly just small little nitpicks from me

SIT.Manager/ViewModels/LinuxSettingsPageViewModel.cs
jubiman: I will be adding a way for users to select which version to use:
image
I have not properly implemented this as I think it is less important, but still nice to add in the future

SIT.Manager/ViewModels/LinuxSettingsPageViewModel.cs
jubiman: Part of the custom environment table structure like in lutris
image

SIT.Manager/Services/Install/InstallerService.cs
```diff
@@ -340,11 +340,16 @@ public Process CreatePatcherProcess(string patcherPath)
         };
         if (OperatingSystem.IsLinux())
         {
-            patcherProcess.StartInfo.FileName = _configService.Config.WineRunner;
+            // TODO: actually improve this (will probably be done after fixing launching problems)
+            LinuxConfig config = (LinuxConfig) _configService.Config;
+            string winePrefix = Path.GetFullPath(config.WinePrefix);
+            // Update the wine prefix and install any required components
+            UpdateWinePrefix(winePrefix);
```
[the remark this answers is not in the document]
jubiman: I'll add some frontend to this to show what is happening, but I won't ask the user whether they want to update the wine prefix as without proper components it will result in improper rendering.

SIT.Manager/Controls/EmbeddedProcessWindow.cs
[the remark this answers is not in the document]
artehe: You're both right.... I realised I made a small oopsie....

https://github.com/stayintarkov/SIT.Manager.Avalonia/blob/4238bf66ca0f75f86b4ed2cb7f7afa121996167d/SIT.Manager/ViewModels/Installation/PatchViewModel.cs#L99

The above was supposed to be ```await embeddedProcessWindow.WaitForExit();```

Probably what I deserve for not testing things properly on Linux 😓 


The whole patching thing is something to look into in the future cause it's a pain on Windows as well as Linux, although Linux is far worse.

SIT.Manager/Services/ManagerConfigService.cs
artehe: I do agree.

Would a far simpler option not be to just create a class called something like LinuxWineConfig and then just add that as a property to the ManagerConfig so that we don't have to handle two different types of manager config?

SIT.Manager/Views/Installation/SelectView.axaml
```diff
@@ -96,6 +97,25 @@
...
+									<MultiBinding Converter="{x:Static BoolConverters.And}">
+										<Binding Path="CurrentInstallProcessState.EftInstallPath" Converter="{x:Static StringConverters.IsNullOrEmpty}"/>
+										<Binding Path="CurrentInstallProcessState.EftVersion" Converter="{x:Static StringConverters.IsNullOrEmpty}"/>
+									</MultiBinding>
+								</Button.IsEnabled>
+								<Grid ColumnDefinitions="*,*">
+									<ui:SymbolIcon Grid.Column="0" Symbol="Add"/>
+									<TextBlock Grid.Column="1" Text="{DynamicResource SelectViewInstallEFTButtonTitle}" TextTrimming="CharacterEllipsis"/>
+								</Grid>
+							</Button>
```
artehe, line 118: Don't need this as we won't be installing EFT ourselves.

SIT.Manager/Views/MainView.axaml
```diff
@@ -90,6 +90,7 @@
 			</ui:NavigationView.MenuItems>
 
 			<ui:NavigationView.FooterMenuItems>
+				<ui:NavigationViewItem IconSource="Settings" Content="{DynamicResource LinuxSettingsPageTitle}" Tag="SIT.Manager.Views.LinuxSettingsPage" IsVisible="{OnPlatform Default=False, Linux=True}"/>
 				<ui:NavigationViewItem IconSource="Settings" Content="{DynamicResource SettingsTitle}" Tag="SIT.Manager.Views.SettingsPage"/>
```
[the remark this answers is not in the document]
jubiman: I felt like it would clutter the settings page (especially since the default window size isn't that big).
I also feel like having a separate button makes it a bit easier to find for the user so they know they might want to check it out, instead of quickly scrolling through the otherwise long list of settings.
Merging the linux settings page back into the settings shouldn't be too difficult, however, so if you'd prefer that. The call is yours to make.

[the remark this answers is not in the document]
artehe: You're right it would make it far too cluttered, I didn't explain what I meant to well. 

Excuse the roughness of these two examples I just did them very quickly but should hopefully show what I was meaning:

Either something along the lines of thise:
image


Or what I think would work better is splitting the whole settings page out into something more like this:
image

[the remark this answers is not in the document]
jubiman: That looks very slick indeed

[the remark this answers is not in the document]
artehe: The more I look at it the more I like this one, it just needs a bit of work to make it look properly nice but I think that'd work well: 
322460309-d296852c-561c-4e39-8449-ba284cfe8efd

You can do it if you want but we can just leave the way you've done the Linux settings page for now and can change all the settings to look this way in a separate PR.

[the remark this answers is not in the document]
jubiman: I'm not the best front-end guy so I think it's better to leave that to someone else :sweat_smile:

artehe:
So sorry I caused a minor merge conflict sorting some stuff from the last couple of days, so fix that and then I think it's good to merge 😄 

Hopefully Linux will work a bit better for people then. Don't worry about the settings for now I'll put them all nicely into one page.

Plootie (changes requested):
Just something I spotted while checking the status of this PR. Would you be able to add the localization keys to the other localizations too? You don't need to translate them, just copy and paste the english keys into the other files (I do plan on changing how this works at some point)

Plootie (approved):

PR #188, opened 2024-04-16
Added localization missing keys and fixed infinite loop of update checking.

J0nathan550:


SIT.Manager/Services/AppUpdaterService.cs
```diff
@@ -100,7 +100,7 @@ public async Task<bool> CheckForUpdate()
         Version currentVersion = Assembly.GetEntryAssembly()?.GetName().Version ?? new Version("0");
         Version latestVersion = new();
 
-        if (_managerConfigService.Config.LookForUpdates && DateTime.Now.AddHours(-1) < _managerConfigService.Config.LastManagerUpdateCheckTime)
+        if (_managerConfigService.Config.LookForUpdates && DateTime.Now.AddHours(1) < _managerConfigService.Config.LastManagerUpdateCheckTime)
```
artehe, line 103: This has hurt my head so much every time I've looked at it.

I think this is now the right way round? Well done for working it out.

PR #189, opened 2024-04-16
reimplemented check update condition, fixed SIT update check.

J0nathan550:
Also fixed the typo.


J0nathan550:
Do not approve now still issues.

artehe (approved):

PR #194, opened 2024-04-19
Update settings page

artehe:
Putting this up now to check what people think of the new look added a screenshots below for reference. Also to see if there's any changes to the way any current settings work/appear which I should add to this. I'm open to any and all suggestions

New view:
image

Old Settings view:
image

Some things that still need doing
- [x] Fix Language selection not being filled by default
- [x] Headings all having settings in them
- [x] Some alignment on the linux view and others.
- [x] Make the SPT-AKI version selectable


Plootie (approved):
Looks amazing and I found no bugs. Everything seems to be working as intended :D

PR #197, opened 2024-04-24
Implement pulling side icons from server

Plootie:
The icons are a little crispy for some reason. I assume it's due to how the bitmaps are being scaled but I can address that at a later date

### DIFF
--- a/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
+++ b/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
@@ -1,4 +1,5 @@
 ﻿using Avalonia.Media.Imaging;
+using Avalonia.Styling;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
 using FluentAvalonia.UI.Controls;
@@ -29,7 +30,7 @@ public partial class CharacterSummaryViewModel : ObservableRecipient
     private AkiMiniProfile _profile;
 
     [ObservableProperty]
-    private Bitmap _sideImage;
+    private Bitmap? _sideImage;
 
     [ObservableProperty]
     private double _levelProgressPercentage = 0;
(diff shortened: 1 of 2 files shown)


SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
```diff
@@ -52,6 +68,20 @@ public CharacterSummaryViewModel(AkiServer server, AkiMiniProfile profile, ILoca
...
+            if(!string.IsNullOrEmpty(profile.Side) && !profile.Side.Equals("unknown", StringComparison.InvariantCultureIgnoreCase))
+            {
+                string cacheKey = $"side_{Profile.Side} icon";
+                CacheValue<Bitmap> cacheVal = await _cachingService.InMemory.GetOrComputeAsync<Bitmap>(cacheKey, async (key) =>
+                {
+                    return new(await _akiServerRequestingService.GetAkiServerImage(server, $"launcher/side_{profile.Side.ToLower()}.png"));
+                });
+                if(cacheVal.HasValue && cacheVal.Value != null)
+                    SideImage = cacheVal.Value;
+            }
```
artehe, line 82: Would this make more sense to be a function that you just await in the Task run?

SIT.Manager/Views/Play/CharacterSummaryView.axaml
```diff
@@ -11,9 +11,11 @@
...
 			Background="Black">
 		<Grid ColumnDefinitions="Auto, *, Auto">
-			<TextBlock Grid.Column="0"
-					   VerticalAlignment="Center"
-					   Text="{Binding Profile.Side}"/>
+			<Image Grid.Column="0"
+					VerticalAlignment="Center"
+					Source="{Binding SideImage}"
+					MaxWidth="80"
+					MaxHeight="80"/>
```
artehe, line 18: If you set Stetch="Uniform" then I think that might help the images be clearer

Plootie: I *might* have slightly copied SPT's launcher and it looks mildly better? Idk exactly why their scaling looks smother compared to ours

artehe (approved):
All good from me, we can fix the todos later with all the others.

PR #198, opened 2024-04-24
fix server logs selection in diagnostics.zip

belettee:
oldest logs were attached to the zip, woopsie daisy


Plootie:
I am not the smartest sometimes...

PR #200, opened 2024-04-24
Play flow fixes

artehe:
Improve the implementation of the play pages flow based on some feelback from @belettee 

- The login dialog should make more sense and not say username twice
- Remember me should now actually remember
- Add character button will register a new character with the server
- Change the start to use a more subtle system that I think might make more sense but we shall see
- The edit server button will be slightly clearer on what it does.


Plootie:
Looks like you've fixed the desync :D just a minor thing, Updating the port number doesn't apply until restarting the manager

Plootie (approved):
Nice work :D

PR #201, opened 2024-04-25
Prevent crash on launching with no internet

artehe:
Add an error on the server install page if there is no version available to install. Also prevent a crash from occuring if the user navigates to the SIT install page with no internet.

### DIFF
--- a/SIT.Manager/ViewModels/MainViewModel.cs
+++ b/SIT.Manager/ViewModels/MainViewModel.cs
@@ -69,7 +69,7 @@ public partial class MainViewModel : ObservableRecipient, IRecipient<Installatio
 
         _localizationService.Translate(new CultureInfo(_managerConfigService.Config.CurrentLanguageSelected));
 
-        var faTheme = Application.Current?.Styles.OfType<FluentAvaloniaTheme>().FirstOrDefault();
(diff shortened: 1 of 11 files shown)


SIT.Manager/ViewModels/MainViewModel.cs
```diff
@@ -87,8 +87,15 @@ private void ManagerConfigService_ConfigChanged(object? sender, ManagerConfig e)
...
-        UpdateAvailable = await _appUpdaterService.CheckForUpdate();
-        SitUpdateAvailable = await _installerService.IsSitUpdateAvailable();
+        try
+        {
+            UpdateAvailable = await _appUpdaterService.CheckForUpdate();
+            SitUpdateAvailable = await _installerService.IsSitUpdateAvailable();
+        }
+        catch (Exception)
+        {
+
```
Plootie, line 97: Maybe put some logger here

Plootie (commented):
Other than the logging comment lgtm

SIT.Manager/ViewModels/MainViewModel.cs
artehe: Oopsie, meant to add that just got distracted. Should be done now 😄

Plootie (approved):

PR #202, opened 2024-04-25
Make EFT Path in settings clearer

artehe:
Should hopefully make the EFT path setting much clearer, and also harder to select the BSG EFT install by mistake now.

I think this is the nicest way of displaying it and doing it but am open to any suggestions.

image

### DIFF
--- a/SIT.Manager/ViewModels/Settings/EftViewModel.cs
+++ b/SIT.Manager/ViewModels/Settings/EftViewModel.cs
@@ -1,5 +1,8 @@
-﻿using CommunityToolkit.Mvvm.Input;
+﻿using CommunityToolkit.Mvvm.ComponentModel;
+using CommunityToolkit.Mvvm.Input;
+using FluentAvalonia.UI.Controls;
 using SIT.Manager.Interfaces;
+using System.IO;
 using System.Threading.Tasks;
 
 namespace SIT.Manager.ViewModels.Settings;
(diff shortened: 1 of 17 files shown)


SIT.Manager/ViewModels/Settings/EftViewModel.cs
```diff
@@ -1,27 +1,46 @@
...
         _barNotificationService = barNotificationService;
+        _installerService = installerService;
         _localizationService = localizationService;
         _versionService = versionService;
 
+        BsgEftInstallPath = Path.GetDirectoryName(_installerService.GetEFTInstallPath()) ?? "Unable to detect BSG EFT install...";
+        SitEftInstallPath = _configsService.Config.SitEftInstallPath;
+        if (string.IsNullOrEmpty(SitEftInstallPath))
+        {
+            SitEftInstallPath = "SIT install not found. Please point to existing install here, or install by clicking the install button on the left";
```
Plootie, line 41: Should this be localized?

Plootie (commented):
Just confirmation on my comment, otherwise looks good

SIT.Manager/ViewModels/Settings/EftViewModel.cs
artehe: oh ye, there's probably a couple other places in this I might have missed 😓

artehe: Should have got all of them now.

Plootie (approved):

PR #214, opened 2024-04-27
More improvements

artehe:
Hopefully this'll detect if you try and install SIT into an existing SPT install and stop you.

Also fix an issue in the OnDisk caching and then use it to remember the last used connection type on the play page to load that when you next startup the launcher

### DIFF
--- a/SIT.Manager/ViewModels/PlayPageViewModel.cs
+++ b/SIT.Manager/ViewModels/PlayPageViewModel.cs
@@ -1,8 +1,10 @@
 ﻿using Avalonia.Controls;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Messaging;
+using SIT.Manager.Interfaces;
 using SIT.Manager.Models.Aki;
 using SIT.Manager.Models.Play;
+using SIT.Manager.Services.Caching;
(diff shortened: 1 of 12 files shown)


Plootie (approved):

SIT.Manager/ViewModels/PlayPageViewModel.cs
```diff
@@ -13,16 +15,39 @@ public partial class PlayPageViewModel : ObservableRecipient,
...
 
+    partial void OnSelectedTabIndexChanged(int value)
+    {
+        if (_cachingService.OnDisk.Exists(SELECTED_TAB_INDEX_CACHE_KEY))
+        {
+            _cachingService.OnDisk.Remove(SELECTED_TAB_INDEX_CACHE_KEY);
+        }
+        _cachingService.OnDisk.Add(SELECTED_TAB_INDEX_CACHE_KEY, value);
+    }
+
```
Plootie, line 50: Sorry about this. I'll add a TryRemove soon. This could cause issues in multithreaded situations

artehe: An update value function would have worked well here too I think, if that's not too much effort.

PR #217, opened 2024-04-28
Add Launch Params

### LINKED ISSUE #170: launch params?
From Discord: [Message Link](https://discord.com/channels/1175114933713776690/1226186259559350273/1226369024296878121d)

> Would be nice to have launch params to the manager to direct connect to a specific ip.
> 
> SPT has already something like this: https://hub.sp-tarkov.com/doc/entry/12-tutorial-play-with-steam-overlay/
> 
> Use case would be to create a shortcut with those launch options to just double click and directly boot into tarkov.
> Or to add it to steam with those launch options to gain access to the steam overlay like in the SPT tutorial.

Neuheit:
Closes #170 

This PR adds the launch params described in the issue.

--address (-a): sets the server address to connect to, if omitted this will start a local server
--username (-u): sets the username of the account to login with
--password (-p): sets the password of the account to login with

### DIFF
--- a/SIT.Manager/ViewModels/Play/DirectConnectViewModel.cs
+++ b/SIT.Manager/ViewModels/Play/DirectConnectViewModel.cs
@@ -75,6 +75,28 @@ public partial class DirectConnectViewModel : ObservableRecipient
         QuickPlayCommand = new AsyncRelayCommand(async () => await ConnectToServer(true));
     }
 
+    public Task ConnectToServer(string address, string username, string password) 
+    {
+        if(!string.IsNullOrEmpty(address))
+        {
+            LastServer = address;
+        }
+
+        if(!string.IsNullOrEmpty(username))
+        {
+            Username = username;
+        }
+
+        if(!string.IsNullOrEmpty(password))
+        {
+            Password = password;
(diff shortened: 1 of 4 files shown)


SIT.Manager/ViewModels/Play/DirectConnectViewModel.cs
```diff
@@ -75,6 +75,28 @@ public partial class DirectConnectViewModel : ObservableRecipient
         QuickPlayCommand = new AsyncRelayCommand(async () => await ConnectToServer(true));
     }
 
+    public Task ConnectToServer(string address, string username, string password) 
+    {
+        if(!string.IsNullOrEmpty(address))
+        {
+            LastServer = address;
+        }
```
Plootie, line 83: If the address is null you'll need to set it to 127.0.0.1 otherwise it'll fail the address validation rule and never start the local server

Plootie (changes requested):
Hey. Thanks for doing this, saved me doing it :D
I Just have this one change to request but other than that it all works great :3

Plootie (approved):
Works great :3

PR #223, opened 2024-04-28
Text update

artehe:
Update the text on the install selection page to make it clearer what is happening on that page, also allow an option to skip the compatibility check on mods by using '*'.


Plootie (approved):
Good call

PR #231, opened 2024-04-29
Updated UI and fix a few bugs in the Manager.

J0nathan550:
Removed also Bender font because it was not really using much in the launcher and everyone hated it.
Also updated missing keys once again.


SIT.Manager/Views/Installation/SelectView.axaml
```diff
@@ -106,16 +103,16 @@
 			</StackPanel>
 
 			<!-- Server -->
-			<StackPanel Grid.Row="1" VerticalAlignment="Center">
+			<StackPanel VerticalAlignment="Center">
 				<TextBlock Text="{DynamicResource SelectViewServerSectionHeading}" TextTrimming="CharacterEllipsis" Classes="FrameHeading"/>
```
artehe, line 107: I'm highlighting it here as well but these textblocks with the headings to the borders can look off at the moment and/or be hard to read.

artehe (changes requested):
In general I think things look better but a lotta changes to make... sorry 😓

SIT.Manager/Views/Installation/SelectView.axaml
J0nathan550: Added shadow effect. You will see it.

artehe (approved):
I'm happy enough with this. 

The headings in places look mostly better there's still one or two places they're not quite right but we can come up with something better for that before we release this again.

The other thing which we can do something to later is the buttons have a strange hover effect sometimes now which flickers mainly on the the install pages, can just sort this in another PR as well.

J0nathan550:
> I'm happy enough with this.
> 
> The headings in places look mostly better there's still one or two places they're not quite right but we can come up with something better for that before we release this again.
> 
> The other thing which we can do something to later is the buttons have a strange hover effect sometimes now which flickers mainly on the the install pages, can just sort this in another PR as well.

I wish to know how to animate stuff in Avalonia. I would do it in seconds. But it's a tricky thing.

artehe:
> > I'm happy enough with this.
> > The headings in places look mostly better there's still one or two places they're not quite right but we can come up with something better for that before we release this again.
> > The other thing which we can do something to later is the buttons have a strange hover effect sometimes now which flickers mainly on the the install pages, can just sort this in another PR as well.
> 
> I wish to know how to animate stuff in Avalonia. I would do it in seconds. But it's a tricky thing.

There was something somewhere which explained it rather well, but there's also this https://docs.avaloniaui.net/docs/guides/graphics-and-animation/keyframe-animations which should explain it as well.

PR #236, opened 2024-05-01
Add port checker tool

Plootie:


Plootie:
Not ready for merge (That will be very obvious from looking at it ;P) But just wanting a general review to make sure I'm not committing to some horrible choices that'll need changed later

SIT.Manager/Views/MainView.axaml
```diff
@@ -26,7 +26,7 @@
 			</ui:TeachingTip.IconSource>
 		</ui:TeachingTip>
 
-		<StackPanel Name="InfoBarStackPanel" Grid.Row="0" x:FieldModifier="Public">
+		<StackPanel Name="InfoBarStackPanel" Grid.Row="0" x:FieldModifier="public">
```
artehe, line 29: To be honest I don't think it even needs  x:FieldModifier so could just remove it.

SIT.Manager/Views/Tools/NetworkToolsView.axaml.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace SIT.Manager.Views.Tools;
+public partial class NetworkToolsView : ActivatableUserControl
+{
+    private readonly NetworkToolsViewModel _dc;
+    public NetworkToolsView()
+    {
+        InitializeComponent();
+        _dc = App.Current.Services.GetService<NetworkToolsViewModel>() ?? throw new Exception("meow");
+        this.DataContext = _dc;
```
artehe, line 16: If you use GetRequiredService here instead then it won't return a nullable and throw and exception instead just like you're doing here.

SIT.Manager/Views/Tools/NetworkToolsView.axaml.cs
```diff
@@ -0,0 +1,31 @@
...
+        base.OnActivated();
+        _dc.RequestCancellationSource = new CancellationTokenSource();
+    }
+
+    protected override void OnDeactivated()
+    {
+        base.OnDeactivated();
+        _dc.RequestCancellationSource.Cancel();
+        _dc.RequestCancellationSource.Dispose();
+    }
```
artehe, line 30: You can move both of these into the view model, there's an OnActivated and OnDeactivated method available if you inherit from ObservableRecipient instead of ObservableObject.

SIT.Manager/Models/Tools/PortCheckerResponse.cs
```diff
@@ -0,0 +1,22 @@
...
+    [JsonPropertyName("portsUsed")] public PortCheckerPorts PortsUsed { get; init; } = new();
+
+    [JsonPropertyName("ipAddress")] public string? IpAddress { get; init; } = "x.x.x.x";
+}
+
+public class PortCheckerPorts
+{
+    [JsonPropertyName("akiPort")] public string AkiPort { get; set; } = "6969";
+    [JsonPropertyName("relayPort")] public string RelayPort { get; set; } = "6970";
+    [JsonPropertyName("natPort")] public string NatPort { get; set; } = "6971";
```
artehe, line 21: Am I being stupid or is there not 4 ports, 6972 as well right? Or does that one not really matter?

SIT.Manager/App.axaml.cs
```diff
@@ -161,33 +167,55 @@ private static ServiceProvider ConfigureServices
                 MaxConnectionsPerServer = 10
             });
 
-            services.AddResiliencePipeline<string, HttpResponseMessage>("default-pipeline", builder =>
+            services.AddHttpClient<NetworkToolsViewModel>(client =>
             {
-                builder.AddRetry(new RetryStrategyOptions<HttpResponseMessage>()
+                client.BaseAddress = new Uri("http://backend.sitcoop.org/");
```
artehe, line 172: Should this string be a const at the top so it's easier to find and change in the future, assuming that it is the address we are meant to be using?

.gitignore
```diff
@@ -396,3 +396,9 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
+/.idea/.idea.SIT.Manager/.idea/.gitignore
+/.idea/.idea.SIT.Manager/.idea/.name
+/.idea/.idea.SIT.Manager/.idea/avalonia.xml
+/.idea/.idea.SIT.Manager/.idea/discord.xml
+/.idea/.idea.SIT.Manager/.idea/indexLayout.xml
+/.idea/.idea.SIT.Manager/.idea/vcs.xml
```
artehe, line 404: Might want to just exclude the whole ```/.idea``` folder as there might be a whole load more files it generates in the future?

artehe (commented):
Only a couple comments so far. Although I haven't run it yet, will wait for you to be done before I do that.

.gitignore
Plootie: Oh, thanks rider for generating the worst possible gitignore changes. Yeah I'll change that :3

SIT.Manager/Models/Tools/PortCheckerResponse.cs
Plootie: 6972 is used but it's a little weird. I don't know the exacts around it but it's just for stuff like insurance and flea market messages and it always seems to show up closed on port tests. There might be a way to test it but that's beyond my knowledge so we'd need to ask bullet or one of the others

.gitignore
artehe: Hehe I wasn't sure if they were auto generated or not 🤷

PR #238, opened 2024-05-01
Change the navigation menu to use something other than NavigationView 

artehe:
The fluent nav menu kept annoying me and had weird bugs to do with using the back button on your mouse that you couldn't disable? So have changed it to something simpler using a split view, in general this seems to work as well, Also makes another change I wanted to make easier adding in the Stay In Tarkov image into the empty space in the menu, so for comparison i've put the two manager UIs below:

Current Manager
image

Updated Manager:
image

I'd love some more suggestions on how to improve the look of the launcher as well


Plootie:
Damn that actually looks much better! We could maybe experiment with transparency? Haven't tested yet but something like a 60-80% alpha often looks nice. Might not look good in our case but worth a try?

J0nathan550:
I don't like the margin on the menu bar if it is possible to change it can you remove it?
(Talking about this)
Screenshot_3
And can we make in a breadcrumb? and extend view to title bar to make it more fluent?

J0nathan550:
We can make something like this I made while ago in my "SoundScapes" project. here is the video of what I am talking about.

https://github.com/stayintarkov/SIT.Manager.Avalonia/assets/88881607/28bff76d-357d-420f-b236-b121823793e6

Plootie:
> I don't like the margin on the menu bar if it is possible to change it can you remove it? (Talking about this) Screenshot_3 And can we make in a breadcrumb? and extend view to title bar to make it more fluent?

Personally I like moving the text away from the edge. Maybe reduce the margin by a few pixels instead?

artehe:
> > I don't like the margin on the menu bar if it is possible to change it can you remove it? (Talking about this) Screenshot_3 And can we make in a breadcrumb? and extend view to title bar to make it more fluent?
> 
> Personally I like moving the text away from the edge. Maybe reduce the margin by a few pixels instead?

I agree have reduced the margin by half, as I do agree it was a little bit far away.

artehe:
> We can make something like this I made while ago in my "SoundScapes" project. here is the video of what I am talking about.
> 2024-05-02.12-56-49.mp4

I would be tempted to do something like this but as you can't extend into the title bar on Linux I feel like this would just look really off there.

artehe:
> if you dont mind, i might chime in with a minor request, can we set the SPT-AKI to AKI-SERVER? the way it is now, creates confusion
> 
> thanks

I went with replacing all instances of SPT-AKI with AKI server 😄

J0nathan550:
> > if you dont mind, i might chime in with a minor request, can we set the SPT-AKI to AKI-SERVER? the way it is now, creates confusion
> > 
> > thanks
> 
> I went with replacing all instances of SPT-AKI with AKI server, in the english version at least 😄 
> 
> @J0nathan550 would you mind doing the same for the other languages please?

Sure

SIT.Manager/Views/MainView.axaml
```diff
@@ -51,98 +66,70 @@
...
-									   Tag="SIT.Manager.Views.PlayPage" 
-									   ToolTip.Tip="{DynamicResource PlayTitleToolTip}"/>
+		<SplitView Background="Black"
+				   IsPaneOpen="True"
+				   DisplayMode="Inline"
+				   OpenPaneLength="166"
+				   Grid.Row="1">
+			<SplitView.Pane>
+				<Grid RowDefinitions="Auto, *, Auto"
+					  Background="Black">
```
Plootie, line 77: I tried out `#080808` and IMO it looks a little nicer. Could give it a try and see how ya feel about it? :3

artehe: okie I went with it.

Plootie (approved):
lgtm :3

PR #253, opened 2024-05-07
Fix strings not localising

### LINKED ISSUE #235: [Bug] Localization in some pages doesn't translate some keys when localization is changed.
There are more pages. This one is just an simple example.

https://github.com/stayintarkov/SIT.Manager.Avalonia/assets/88881607/689fed0e-8fd3-4412-a839-8fc1dc63a624

artehe:
Use a different textblock and just show or hide it as necessary so that it will reload the localised string properly.

Resolve #235

### DIFF
--- a/SIT.Manager/ViewModels/Settings/EftViewModel.cs
+++ b/SIT.Manager/ViewModels/Settings/EftViewModel.cs
@@ -20,6 +20,9 @@ public partial class EftViewModel : SettingsViewModelBase
     [ObservableProperty]
     private string _sitEftInstallPath;
 
+    [ObservableProperty]
+    private bool _showEftInstallPathMissing;
+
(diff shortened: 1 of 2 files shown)


Plootie (commented):
Just curious. Would it be worth using the `StringConverters.IsNullOrEmpty` for the visibility over doing it in the viewmodel? Either way works good so just reminding ya it exists if you wanna use it. If not all good to merge!

artehe:
> Just curious. Would it be worth using the `StringConverters.IsNullOrEmpty` for the visibility over doing it in the viewmodel? Either way works good so just reminding ya it exists if you wanna use it. If not all good to merge!

I knew I was doing something stupid, was literally thinking like I'm not doing this right. I shall sort it.

Plootie (approved):

PR #254, opened 2024-05-07
Move the manager version string to be always visible

artehe:
Moves the manager version string from where it currently is on the settings page as shown below to be always visible but tucked away in the bottom left corner.

Before;
image

After:
image


Plootie (approved):
Good idea :3

PR #257, opened 2024-05-14
Update UI/UX for most pages

artehe:
The main thing I wanted to do here was start making the UI much easier to see in places, starting with the update page. If we like what is happening here I will go through all the other pages and make them all consistent (which I can do in this PR or seperate ones)

Below is a side by side of what the master branch currently looks like and what I'm proposing;
image

Summary of changes:
- Remove some unecessary TODOs as they have been done or not relevant anymore
- Add a subtle curve to the bottom of the navigation window and change it's coloring back to black
- Make updating the manager take 1 less click
- Overhaull the look of most pages, makes other places look a little weird currently.
- Add some visuals to updating for debugging without it actually updating for you, and will always tell you there is an update in debug builds now to make checking things easier.


Plootie (approved):
Looks amazing. Definitely a good direction to start taking the UI

PR #266, opened 2024-05-15
Various Fixes

### LINKED ISSUE #245: Block installing into OneDrive
This is *known* to cause issues and I see a concerning amount of users doing this

artehe:
A few fixes and tweaks to hopefully improve things:

1. Reduce the excessive number of pings to bookmarked servers as it would be pinging the server even when the view wasn't activated
2. Add a first super basic check for installing to OneDrive should resolve #245
3. Fix the manager getting confused and showing multiple text over each other when looking at EFT settings
4. Show a more helpful error message when unable to determine a SIT version to install.

### DIFF
--- a/SIT.Manager/ViewModels/Installation/ConfigureSitViewModel.cs
+++ b/SIT.Manager/ViewModels/Installation/ConfigureSitViewModel.cs
@@ -52,6 +52,12 @@ public partial class ConfigureSitViewModel : InstallationViewModelBase
     [ObservableProperty]
     private bool _hasRecommendedModsAvailable = false;
 
+    [ObservableProperty]
+    private bool _showNoAvailableSitVersionSelectionError = false;
+
+    [ObservableProperty]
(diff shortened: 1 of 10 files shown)


SIT.Manager/ViewModels/Installation/ConfigureSitViewModel.cs
```diff
@@ -168,6 +188,19 @@ private async Task FetchVersionAndMirrorMatrix()
...
+    /// </summary>
+    /// <returns>True if this is a OneDrive directory otherwise false</returns>
+    private static bool HasSelectedOneDriveInstallPath(string requestedDirectory)
+    {
+        if (requestedDirectory.Contains("OneDrive"))
+        {
+            return true;
+        }
+        return false;
+    }
```
Plootie, line 202: `return requestedDirectory.Contains("OneDrive")`? ;P

Plootie (commented):
Small nit, other than that good to merge :3

PR #273, opened 2024-05-17
Caching service tweaks

Plootie:
Idk wtf I was doing when I wrote this the first time but this should hopefully fix whatever the hell was going on when we started to use this


artehe (approved):
I didn't have any problems so am down to try it and see what happens

PR #276, opened 2024-05-17
Make the buttons on the configure installs pages more unified

artehe:
Essentially apply the same theming ideas to these buttons as there is almost everywhere else. It doesn't seem to work brilliantly but I think that's more down to some things in FluentAvalonia getting in the way, can look into those later.


Plootie (approved):
Looking really good :3

PR #278, opened 2024-05-17
Fix stream disposing in return

Plootie:


artehe (approved):
still works so I'm happy :)

PR #283, opened 2024-05-21
Rework Mods Page

### LINKED ISSUE #232: Change the mod page
After the implementation of native mod support unless someone has a good idea how to do it that doesn't involve scraping the SPT website then I propose that the mod page should change to be more something just to manage the mods that are installed. Below is a list of things I think the new page should do:

- [ ] Show the currently installed mods and their versions/info
- [ ] Allow for enabling and disabling of mods

artehe:
Putting this up now in case people have some thoughts or suggestions as this will resolve #232

In short this converts the mods page from the current implementation we have to one which is more of a management of mods, in the sense that it won't install or uninstall mods for you but give you the option to enable or disable them. Also this will list all the current mods active for your EFT install and their versions.

The other part as I think this makes sense will be giving the ability to download the mod compatibility layer, but is kept seperate from the main SIT install.

Main things still to do are listed in the issue, but is essentially just the actual logic which runs the page. An example of what I think the page should look like now would be:

image


artehe:
Lemme know what you guys think 😄

Plootie (commented):
Everything on the code side looks good! Just a few things I noted with the UI. the favorite button on the right could do with a tiny bit of margin to bring it out of the scroll bar and if possible could the version column be stopped from hogging space the name column could use?

artehe:
> Everything on the code side looks good! Just a few things I noted with the UI. the favorite button on the right could do with a tiny bit of margin to bring it out of the scroll bar and if possible could the version column be stopped from hogging space the name column could use?

Ah, my stupidity is coming back to bite me. I should replace the star/favourite with something more obvious that it's meant to enable and/or disable the mods. Don't know why I didn't just use something like a toggle button here 🫤 

I'll change out the star for something more obvious then add the margin and restrict the version columns sizing.

Plootie (approved):
All lgtm :3

PR #284, opened 2024-05-22
Added new setting to allow the launcher to Minimize on game launch - …

stevejackson80:
…tested in Desktop mode (should only effect Desktop mode - please correct me if I am wrong, never used Avalonia before).

Language resources updated with text for setting and tooltip - can only test en-US, other languages were done with online translating tools - however text looks very similar to the setting for Close on launch so presume it is ok.  Really needs to be checked by native language speakers.

Resulting launcher settings screen is as follows:

image

As soon as the Play button is pressed (or successful login), launcher checks to see if the minimize setting is enabled, that we are in desktop mode, and that there is a MainWindow - if so WindowState is set to Minimized as the EFT executable is started..


Plootie (approved):
lgtm ^^

PR #285, opened 2024-05-23
Enable and Disable the Login and Play buttons on Character Summary view dependent on EFT's running state.

stevejackson80:
Updated the CharacterSummaryViewModel to expose an additional property _canLaunch (with an Observable generated instance) which is set to either true or false (defaults to true) depending on the state of _tarkovClientService.RunningStateChanged.

An event handler is attached to listen for RunningStateChanged events, if the state is either Starting or Running _canLaunch is set to false, the other states (NotRunning and StoppedUnexpectedly) set this to true.

The CharacterSummaryView.axaml's command button (Login or Play) has the property IsEnabled bound to this new property - in doing this all buttons are enabled or disabled when EFT is started/stopped. This property is changed on the UI thread with the use of Dispatcher.UIThread.Invoke().

The upshot of this change is that the UI now behaves as such:

Not logged in, Tarkov not running, all buttons enabled:
Unsaved profile - Tarkov not launched

Login dialog disables the login button for the selected profile (as expected for a button with a bound command):
Unsaved profile - Login Dialog

Successful login, Tarkov now running (or starting up), all buttons are disabled:
Unsaved profile - Tarkov running

Exited from Tarkov, all buttons are enabled.
Unsaved profile - Tarkov exited

Saved profiles work in much the same way - but without the login dialog of course.

Saved profile (remember me ticked during login), Tarkov not running all buttons available.
Saved profile - Tarkov not launched

Saved profile (remember me ticked during login), Tarkov running all buttons disabled.
Saved profile - Tarkov launched

Saved profile (remember me ticked during login), Tarkov exited all buttons available again.
Saved profile - Tarkov exited

> [!NOTE]
> Have added comments in the CharacterSummaryViewModel constructor regarding future implementation of IActivatableViewModel from Avalonia.ReactiveUI - however it would require changes to the code behind the axaml. It *may* be worth looking into this as part of a larger refactor at a later date.

### DIFF
--- a/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
+++ b/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
@@ -1,4 +1,5 @@
 ﻿using Avalonia.Media.Imaging;
+using Avalonia.Threading;
 using CommunityToolkit.Mvvm.ComponentModel;
 using CommunityToolkit.Mvvm.Input;
 using FluentAvalonia.UI.Controls;
@@ -40,6 +41,9 @@ public partial class CharacterSummaryViewModel : ObservableRecipient
 
     [ObservableProperty]
     private bool _requireLogin = true;
+
+    [ObservableProperty]
+    public bool _canLaunch = true;
 
     public IAsyncRelayCommand PlayCommand { get; }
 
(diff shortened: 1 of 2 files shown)


Plootie (approved):
also lgtm :3 In the future I want to change this up to some kind of splash screen or something but a good stopgap ^^

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
```diff
@@ -86,6 +90,11 @@ public partial class CharacterSummaryViewModel : ObservableRecipient
         Task.Run(SetSideImage);
 
         PlayCommand = new AsyncRelayCommand(Play);
+
+        // In an ideal world we would use OnActivated and OnDeactivated - which are implemented from IActivatableViewModel in the Avalonia.ReactiveUI package.
+        // However, this would also require changes in the CharacterSummaryView class - for not this implementation, while crude, does suffice.
+        // It may be worth implementing Avalonia.ReactiveUI.IActivatableViewModel at a later date for all pages as part of a larger refactor.
+        _tarkovClientService.RunningStateChanged += TarkovClient_RunningStateChanged;
```
artehe, line 97: OnActivated & OnDeactivated do exist, they are just there as overrides

PR #286, opened 2024-05-23
Added a logout button so that a "saved profile" can be forgotten.

stevejackson80:
Useful if the PC is shared or if you, for whatever reason, need to share your SIT install.

New button only displays if the profile is a saved profile and updated localization files (using online translate tools) to add button text.

Character selection with a saved profile (remember me ticked on login dialog):

Saved profile - logout button

After clicking on logout:

Saved profile - logged out

ManagerConfig.json file is updated to remove the saved profile after clicking so that the next time the manager is started the profile is still logged out.

### DIFF
--- a/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
+++ b/SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
@@ -145,14 +145,6 @@ public partial class CharacterSummaryViewModel : ObservableRecipient
             if (success && rememberLogin)
             {
                 _connectedServer.Characters.Add(character);
-
-                int index = _configService.Config.BookmarkedServers.FindIndex(x => x.Address == _connectedServer.Address);
-
-                if (index != -1 && !_configService.Config.BookmarkedServers[index].Characters.Any(x => x.Username == character.Username))
-                {
-                    _configService.Config.BookmarkedServers[index].Characters.Add(character);
-                }
-
                 _configService.UpdateConfig(_configService.Config);
                 RequireLogin = false;
             }
@@ -172,18 +164,9 @@ public partial class CharacterSummaryViewModel : ObservableRecipient
 
     private async Task Logout()
     {
-        AkiCharacter? character = _connectedServer.Characters.FirstOrDefault(x => x.Username == Profile.Username);
-
         if (character != null)
         {
             _connectedServer.Characters.Remove(character);
-
-            int index = _configService.Config.BookmarkedServers.FindIndex(x => x.Address == _connectedServer.Address);
-
-            if (index != -1 && !_configService.Config.BookmarkedServers[index].Characters.Any(x => x.Username == character.Username))
-            {
-                _configService.Config.BookmarkedServers[index].Characters.Remove(character);
-            }
             _configService.UpdateConfig(_configService.Config);
             RequireLogin = true;
         }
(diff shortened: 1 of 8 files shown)


SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
```diff
@@ -161,4 +169,23 @@ private async Task Play()
             await errorDialog.ShowAsync();
         }
     }
+
+    private async Task Logout()
+    {
+        AkiCharacter? character = _connectedServer.Characters.FirstOrDefault(x => x.Username == Profile.Username);
```
Plootie, line 175: We already keep an `AkiCharacter` reference in the class that you could use if I'm not mistaken

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
```diff
@@ -161,4 +169,23 @@ private async Task Play()
...
+        if (character != null)
+        {
+            _connectedServer.Characters.Remove(character);
+
+            int index = _configService.Config.BookmarkedServers.FindIndex(x => x.Address == _connectedServer.Address);
+
+            if (index != -1 && !_configService.Config.BookmarkedServers[index].Characters.Any(x => x.Username == character.Username))
+            {
+                _configService.Config.BookmarkedServers[index].Characters.Remove(character);
+            }
```
Plootie, line 186: I may also be wrong about this but since `_connectedServer` is passed to us as a reference, the characters list there should be the same character list that's referred to here

Plootie (changes requested):

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
stevejackson80: Your right, I even missed ReSharper telling me that my new object was hiding an existing one DOH!  I'll get that sorted (and tested) shortly.

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
stevejackson80: I actually based this code on the code that already saved the login details (assuming one had ticked remember me) - a literal copy/paste (but with Remove rather than Add of course).  But I'll check this too and make any required changes.

stevejackson80: You are 100% bang on.  So I am changing the code to:

```
_connectedServer.Characters.Remove(character);
_configService.UpdateConfig(_configService.Config);
```

The same can also be done for the save login code (line 147 onwards), I've tested it and what is there can be removed and replaced with:

```
_connectedServer.Characters.Add(character);
_configService.UpdateConfig(_configService.Config);
```

Are you happy for this to be optimised too or would you rather it was left as is?

Plootie: Sure! I'll probably end up forgetting about it if it's left to me ;P

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
stevejackson80: Just pushed a commit to resolve this issue.

SIT.Manager/ViewModels/Play/CharacterSummaryViewModel.cs
stevejackson80: Just pushed a commit to resolve this issue.

Plootie (approved):

PR #294, opened 2024-05-28
Port checker enhancements

### LINKED ISSUE #280: minor changes for portchecker
- add few sec cooldown for the run button
- add option to "ping" external server, from local (can be checked with `TcpClient.ConnectAsync`, if it errors out, the port is not available)

artehe:
Resolve #280 by adding the suggested improvements.


Plootie:
Everything looks great to me, just one small thing. If the remote address doesn't respond instead of rejecting the request the checker will be stuck waiting for 60 seconds. (this is a problem for both modes). Do you think we could chuck a custom timeout on it?

Plootie (approved):
Disregard my previous comment, I'm actually gonna stop dragging my feet and get on with the refactor, as such I'm happy to push this through as is and I can do it myself when I get to this section

PR #295, opened 2024-05-28
Implement server nicknaming locally for users

### LINKED ISSUE #267: Add naming of server functionality
Add the option when installing the server to provide a custom name if you wish which would then show up in the bookmarks section of the manager. 

At the same time could also look to implement the ability to edit the name of the sever from this dialog?
image
Though this could prove a lot more challenging than it's worth.

artehe:
Resolve #267 by adding an optional entry in the edit server and add server dialogs to set a nickname for the server.

This nickname overrides whatever the server name is set to and displays that instead, if a user wants to see the real server name again then they just have to unset the nickname by editing the server entry to remove it.

### DIFF
--- a/SIT.Manager/ViewModels/Play/ServerSummaryViewModel.cs
+++ b/SIT.Manager/ViewModels/Play/ServerSummaryViewModel.cs
@@ -14,7 +14,6 @@ using SIT.Manager.Views.Play;
 using System;
 using System.ComponentModel;
 using System.IO;
-using System.Linq;
 using System.Threading.Tasks;
 
 namespace SIT.Manager.ViewModels.Play;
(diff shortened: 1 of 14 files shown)


SIT.Manager/ViewModels/Play/ServerSummaryViewModel.cs
```diff
@@ -62,10 +62,27 @@ public string ShownAddress
...
+    {
+        get
+        {
+            AkiServer? bookmarkedServer = _configService.Config.BookmarkedServers.FirstOrDefault(x => x.Address == _server.Address);
+            if (bookmarkedServer != null)
+            {
+                return !string.IsNullOrEmpty(bookmarkedServer.Nickname) ? bookmarkedServer.Nickname : Name;
+            }
+            return Name;
+        }
```
Plootie, line 75: We store a self reference to the server as _server which should be the same reference as what this code retrieves

SIT.Manager/ViewModels/Play/ServerSummaryViewModel.cs
```diff
@@ -116,20 +133,22 @@ private void Delete()
 
     private async Task Edit()
     {
-        CreateServerDialogView dialog = new(_localizationService, true, Address.AbsoluteUri);
-        (ContentDialogResult result, Uri serverUri) = await dialog.ShowAsync();
-        if (result == ContentDialogResult.Primary)
+        AkiServer? currentServer = _configService.Config.BookmarkedServers.FirstOrDefault(x => x.Address == _server.Address);
```
Plootie, line 136: Refer to other comment about self reference

Plootie (changes requested):
Just these two, everything else looks and works great!

SIT.Manager/ViewModels/Play/ServerSummaryViewModel.cs
artehe: Fair point, I probably should have just sorted out the problem it caused in the first place 😄

Plootie (approved):

